PR #2, opened 2025-06-10
[Eagle3] feat: Impl TTT feature

FlamingoPg:
Impl ttt framework, but still not tested.


spec/models/llama4/_eagle_attention.py
sleepcoo, line 1: remove it in the modules？ other models can also use it.

PR #5, opened 2025-06-21
add draft and base model

FlamingoPg:


FrankLeeeee:
Can you also implement the `sgl_spec.modeling.auto_model.AutoModelForCausalLM`? We want to use in this way

```python
from sgl_spec import AutoModelForCausalLM

base_model = AutoModelForCausalLM.from_pretrained(...)
draft_model = AutoModelForCausalLM.from_pretrained(...)
```

Do note that we need to keep the "architectures" field in the config file the same as sglang.

You can either do this in this PR or a separate PR.

PR #23, opened 2025-07-13
Support the production and cache reading logic of d2t and t2d. 

sleepcoo:


sgl_spec/data/config.py
```diff
@@ -57,7 +58,7 @@ class DataConfig:
     shuffle_seed: int = 42
     pin_memory: bool = True
     num_workers: int = 4
-    load_from_cache_file: bool = True
+    load_from_cache_file: str = field(default_factory=lambda: str(Path(__file__).resolve().parent.parent / 'cache.pt'))
```
FrankLeeeee: This will cause problems if anyone does not install with the the development mode.

PR #30, opened 2025-07-18
Support VocabParallelEmbedding

sleepcoo:


FrankLeeeee:
If the draft model is in TP mode now, how about the checkpoint saving?

PR #72, opened 2025-07-28
bugfix for skip count in prepare data for ultrachat

wpybtw:
fix bug for skip count in prepare data for ultrachat


sleepcoo:
Why is an extra 0 added to the return value?

sleepcoo:
Great job!!!

PR #80, opened 2025-07-29
add wandb args check

sleepcoo:


specforge/utils.py
```diff
@@ -9,6 +9,28 @@
...
+        args: The parsed arguments
+    """
+    if args.wandb:
+        missing_args = []
+        if args.wandb_project is None:
+            missing_args.append("--wandb-project")
+        if args.wandb_name is None:
+            missing_args.append("--wandb-name")
+        if args.wandb_key is None:
+            missing_args.append("--wandb-key")
```
FrankLeeeee: I don't think all need to be set, key and project are mandatory if I am not wrong. If name is not given, wandb will create a random name. you can test it first.

sleepcoo: > I don't think all need to be set, key and project are mandatory if I am not wrong. If name is not given, wandb will create a random name. you can test it first.

I've confirmed that wandb definitely requires a key, while others can be generated randomly. I've added checks for the wandb key along with other methods to obtain it.

specforge/utils.py
```diff
@@ -9,6 +10,42 @@
 from transformers import PretrainedConfig
 
 
+def validate_wandb_args(parser, args):
+    if args.wandb_key is not None:
+        return
+
+    if "WANDB_LOCAL" in os.environ:
+        args.wandb_key = "LOCAL"
+        return
```
FrankLeeeee: Does `WANDB_LOCAL` requires a value? e.g. `WANDB_LOCAL=1`. If this is the case, the check should be `if os.environ.get('WANDB_LOCAL', '0') == '1'`.

sleepcoo: I deleted this code. I found that wandb doesn't support this variable. It should be a debugging parameter of other open-source training libraries

FrankLeeeee: ok

FrankLeeeee:
LGTM.

PR #91, opened 2025-07-30
[BugFix] Fix calling `dist.get_rank` before `init_distributed` in `validate_wandb_args`

guoyuhong:
After merging [PR #80](https://github.com/sgl-project/SpecForge/pull/80) , I discovered that the scripts `scripts/train_eagle3_offline.py` and `scripts/train_eagle3_online.py` fail to execute. The root cause is that `dist.get_rank()` is being called before `init_distributed()`, leading to initialization errors.
image

### DIFF
--- a/scripts/train_eagle3_offline.py
+++ b/scripts/train_eagle3_offline.py
@@ -83,10 +83,7 @@ def parse_args():
 
     args = parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


FrankLeeeee (approved):

FrankLeeeee:
Thanks!

PR #92, opened 2025-07-30
Support customize ttt length.

Ximingwang-09:
We need to configure different TTT lengths for different scenarios, which also has a impact on training speed.


FrankLeeeee (approved):
LGTM

PR #197, opened 2025-09-01
Support Train Eagle-3 By DeepSpeed

xq25478:
# Support train eagle3 by deepspeed for large model like 72B/235B


sleepcoo:
Is the training speed improved compared to the original implementation?

zyksir:
Now we have sgl online for training large models. we can use sglang as backend to support different models

PR #212, opened 2025-09-04
Pass only `draft_model` into BF16Optimizer

Wokzy:
Target model weights are converted into fp32 and passed to AdamW just for no reason, which causes OOM

### DIFF
--- a/scripts/train_eagle3_online.py
+++ b/scripts/train_eagle3_online.py
@@ -283,6 +283,10 @@ def main():
                 .eval()
                 .cuda()
             )
+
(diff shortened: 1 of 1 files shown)


yubofredwang (commented):
Nice find! I think we should also be explicit by only pass draft_model into BF16Optimizer. What do you think?

Wokzy:
> Nice find! I think we should also be explicit by only pass draft_model into BF16Optimizer. What do you think?

I agree, haven't thought about it before. The last commit should work fine

yubofredwang (approved):

PR #182, opened 2025-08-26
supported think mode

jiapingW:
## Motivation

We found that for inference models such as the qwq-32b and qwen3 series, the input IDs they construct are inconsistent with the expected ones.
Using qwen3 as an example:
```python
    messages = [
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Who are you?"},
        {"role": "assistant", "content": "I am a model."},
        {"role": "user", "content": "What is your name?"},
        {"role": "assistant", "content": "My name is QwQ."},
    ]
    res = tokenizer.apply_chat_template(
        messages,
        tokenize=False,
        add_generation_prompt=False,
    )
    # response will be:
    # <|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\nWho are you?<|im_end|>\n<|im_start|>assistant\nI am a model.<|im_end|>\n<|im_start|>user\nWhat is your name?<|im_end|>\n<|im_start|>assistant\n<think>\n\n</think>\n\nMy name is QwQ.<|im_end|>\n

    res = tokenizer.apply_chat_template(
        messages[:-1],
        tokenize=False,
        add_generation_prompt=True,
    ) + message[-1]["content"] + "<|im_end|>\n"

    # response will be:
    # <|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\nWho are you?<|im_end|>\n<|im_start|>assistant\nI am a model.<|im_end|>\n<|im_start|>user\nWhat is your name?<|im_end|>\n<|im_start|>assistant\nMy name is QwQ.<|im_end|>\n
```
We can see it add a `<think>\n\n</think>\n\n` token at the end of the last response. It corresponds to the case where qwen3 adopts non-think mode. This is also the current implementation, which causes the Eagle model to train tokens that should not be involved in calculating the loss. As for the currently open source qwen3-eagle data such as https://huggingface.co/datasets/Tengyunw/qwen3_8b_eagle3, the data obtained uses the thinking mode. This will cause the training data and the inference data to be misaligned.

Therefore, we introduce an additional parameter for the inference model: **is-think-mode** to determine whether to use the prompt processing related to think mode. Because different models' **apply_chat_template** handles think mode differently, we choose to splice prompt to complete the following mode. Besides, we modified the template in **template.py** to adapt the corresponding think input to different models.

You can use the following code under the processing.py to debug and verify the correctness of the assistant message processed and the corresponding entire conversation.
```python 
[...]

PR #239, opened 2025-09-22
Feature/add sgl online

zyksir:
## Motivation

This PR uses sglang to do online training. It adds an example of training a EAGLE3 model for Llama3.1 7B, which includes:
- a detailed tutorial for how to train a EAGLE3 model for Llama3.1 7B
- generated data by target model
- potentially support for thinking models

## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).


sleepcoo:
data: 300k 
model: qwen3-235B-fp8
hardware: H100*8
**speedup**
  | Hour
-- | --
online |  
sgl_online | 8

jiapingW:
Your implementation is very elegant. I used your code and scripts to train Qwen2.5-7B. I think there are two areas for improvement: 1. Your sgl_online training requires tp-size to be equal to NUM_GPUS, but in previous online training, they can be different. 2. Only draft-micro-batch-size of 1 is supported.

zyksir:
@jiapingW this PR only make it work to use sglang for online training so that we can use sglang for ANY base models. These two things you mentioned did exists and can be improved in the future.
- For small models, TP=8 only hurt the generating througput. Actually hf implementation if no TP is used.
- For bs=1, we can use GA to control the global batch size. The accuracy only depends on the global batch size. (and from my own point, most times we want to use micro-batch-size=1 to make sure OOM won't happen)

jiapingW:
> @jiapingW this PR only make it work to use sglang for online training so that we can use sglang for ANY base models. These two things you mentioned did exists and can be improved in the future.
> 
> * For small models, TP=8 only hurt the generating througput. Actually hf implementation if no TP is used.
> * For bs=1, we can use GA to control the global batch size. The accuracy only depends on the global batch size. (and from my own point, most times we want to use micro-batch-size=1 to make sure OOM won't happen)

Thanks for your answer, I understand your design.

jiapingW:
@zyksir I used the latest version of Specforge to train Qwen3-235b-a22b-instruct-2507. Your **scripts/build_eagle3_dataset_cache.py** will have a question that Qwen3-instruct doesn't  have **\<think>\</think>** token but because of the apply_chat_template impl, your conversation will have **\<think>\</think>**  token in last chat and these tokens will be joined to train. I think this is not in line with the expected results, just like the problem with my previous https://github.com/sgl-project/SpecForge/pull/182 which was not actually solved.

PR #246, opened 2025-10-05
Update server args in bench speedup

ispobock:
## Motivation

- `--enable-ep-moe ` is deprecated in SGLang, use `--ep-size` instead https://github.com/sgl-project/sglang/blob/4cb5a5235e49f08e9f47165fb7f95e35145fce43/python/sglang/srt/server_args.py#L2767
- add `--disable-radix-cache`


FrankLeeeee (approved):

FrankLeeeee:
LGTM.

PR #278, opened 2025-11-06
added abstraction for target model backend

### LINKED ISSUE #255: [Feature] Refactoring multiple target model options into backend
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/SpecForge/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Currently, we have three ways to building a target model:
1. directly from huggingface transformers: this is the easiest and hf provides the fullest model support
2. use customized distributed model: this is to complement HF as HF does not provide distributed version for every model
3. directly from SGLang： this is to utilize the SGLang's existing model implementation to save some  efforts on writing distributed code

While all of them see to have a role to play in terms of the target modeling, they are now used separately via inconsistent APIs, which can cause some confusion and unfriendliness  to users. I think it is time to refactor this part to turn three of them into an unified interface with different backends.

### Design

We can design an interface for target models:

```python

class Eagle3TargetModel:

     def __call__(self, ...):
           # generate the targets for eagle3
           ...

     def set_aux_hidden_layers(self, layer_ids: Optional[ist[int]] = None):
          ...

class HFEagle3TargetModel(Eagle3TargetModel):
     ...

class CustomEagle3TargetModel(Eagle3TargetModel):
     ...

class SGLangEagle3TargetModel(Eagle3TargetModel):
     ...

```

### Existing Problems

It seems that SGLang and HF are currently not compatible with the case where TP > 1 and DP > 1.

### Related resources

_No response_

FrankLeeeee:
## Motivation

To reduce efforts to integrating new models and re-use the existing code, we have introduced the SGLang runner in the codebase, which will use `ModelRunner` in SGLang to generate the target hidden states. However, the API is a bit messy. To resolve this, we should implement a unified interface to unify HuggingFace, Custom Models and SGLang. The differences among these options are listed below:
- SGLang: most favoured backend, it supports multiple models out of box and provides the same precision when we deploy EAGLE3 to SGLang
- HuggingFace: this optional is used mostly when the wanted model does not exist in SGLang modeling, thus we rely on HuggingFace's modeling for training
- Custom: this is used in cases where the user model is closed-source and the architecture is not supported in SGLang or HuggingFace, or the user wants  to apply some additional system optimizations to the target model.

image

With this unified interface, we can better organize our code for readability and usability.

## Modifications

1. Moved all custom models to `specforge.modeling.target.custom_backend`
2. Implemented a `SGLangRunner` in `specforge.model.target.sglang_backend`. This is mainly to patch the SGLang's distributed groups so that we can have DP+TP in the code. Previously, SGLang model can only have TP size = world size, this restricts us in several cases 1) we don't have to have such TP size, 2) The number of attention head is not divisible by TP size 3) we want to run across multiple nodes. Thus, by patching the distributed-related APIs in SGLang, we can have multiple copies of SGLang models. 
3. Rewrite the training script to compose the logic as functions for better readability.

The parallelism scheme:
- Before

image

- After (look at target modes on GPU 2 and 3)

image

## Related Issues

Fixed #255 

## Accuracy Test

I have run Llama3 on ShareGPT for 2 epochs and achieves 2.55 acceptance length on 3-1-4 configuration on MTBench.

## Checklist

- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).

### DIFF
--- a/specforge/modeling/target/eagle3_target_model.py
+++ b/specforge/modeling/target/eagle3_target_model.py
@@ -254,6 +254,9 @@ class SGLangEagle3TargetModel(Eagle3TargetModel):
         else:
             aux_hidden_states_list = None
 
+        # TODO: can we not clear?
+        self.model_runner.req_to_token_pool.clear()
+        self.model_runner.token_to_kv_pool_allocator.clear()
         return logits, aux_hidden_states_list
 
     def _maybe_prepare_mlp_sync_batch(self, batch: ScheduleBatch):


FrankLeeeee:
@jiapingW @zyksir i have added more detailed description for this PR. I will add both of you as the co-authors of this PR as well.

PR #298, opened 2025-11-13
Fix offline training bugs: missing target_head computation, handling vocab mapping, optimizer resume

hanq-moreh:
#### Fix issues in `scripts/train_eagle3_offline.py`

1. **Missing target computation**  
   In `OfflineEagle3Model.forward()`, the target must be computed using `target_head` **before** calling `super().forward()`.

2. **Handle equal vocab sizes**  
   If the draft model and target model have the same vocab size, vocab mapping is unnecessary.  
   Therefore, the default value of `t2d` in `LlamaForCausalLMEagle3` should be `torch.ones`.

3. **Fix FSDP2 resume loading**  
   When resuming offline training with FSDP2, while model params are `DTensor` but loaded optimizer states are normal `Tensor`, causing errors.  
   After loading, optimizer states must be converted to Distributed Tensors.

## Modifications

1. `scripts/train_eagle3_offline.py`  
   - Skip vocab mapping when draft and target vocab sizes are same.  
   - Fix resume logic so optimizer state loads correctly as distributed tensors.

2. `specforge/core/eagle3.py`  
   - Add missing `target_head` computation in `OfflineEagle3Model.forward()`.

3. `specforge/modeling/draft/llama3_eagle.py`  
   - Change default `t2d` to `torch.ones`.

4. `specforge/utils.py`  
   - Add a helper function to convert optimizer states into Distributed Tensors.

## Checklist

### DIFF
--- a/specforge/modeling/draft/llama3_eagle.py
+++ b/specforge/modeling/draft/llama3_eagle.py
@@ -1012,7 +1012,7 @@ class LlamaForCausalLMEagle3(Eagle3DraftModel):
         )
 
         # create vocab buffers
-        t2d = torch.zeros(self.vocab_size, dtype=torch.bool)
+        t2d = torch.ones(self.vocab_size, dtype=torch.bool)
         d2t = torch.zeros(self.draft_vocab_size, dtype=torch.int64)
         self.register_buffer("t2d", t2d)
         self.register_buffer("d2t", d2t)
--- a/specforge/utils.py
+++ b/specforge/utils.py
@@ -7,6 +7,7 @@ from datetime import timedelta
 
 import torch
 import torch.distributed as dist
+from torch.distributed._tensor import DTensor, Shard, distribute_tensor
 from transformers import AutoConfig, PretrainedConfig
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


specforge/core/eagle3.py
```diff
@@ -264,6 +264,7 @@ def forward(
             vlosses: List of validation losses (not used in this implementation).
             acces: List of accuracies for each TTT step.
         """
+        target = self.target_head(target)
```
FrankLeeeee: this should be decorated with `torch.no_grad()`.

specforge/modeling/draft/llama3_eagle.py
```diff
@@ -1012,7 +1012,7 @@ def __init__(self, config, quant_config=None, attention_backend="sdpa") -> None:
         )
 
         # create vocab buffers
-        t2d = torch.zeros(self.vocab_size, dtype=torch.bool)
+        t2d = torch.ones(self.vocab_size, dtype=torch.bool)
```
FrankLeeeee, line 1015: Why is this ones?

specforge/utils.py
```diff
@@ -248,3 +249,41 @@ def get_full_optimizer_state(optimizer_state_dict: dict):
             for param_id, param_state in optimizer_state_dict["state"].items()
         }
     return full_optimizer_state_dict
+
+
+def shard_optimizer_state_with_dtensor(bf16_optimizer, device_mesh):
```
FrankLeeeee, line 254: Can we just use FSDP's shard_op?

specforge/modeling/draft/llama3_eagle.py
hanq-moreh: When the target vocab size and the draft vocab size are identical, we effectively have a 1-to-1 mapping between the two vocabularies. In this case, we don’t need to compute a vocab mapping at all — we can simply use `torch.ones() `as the default mapping.

If we keep `torch.zeros()` as the default value, the last dimension of the `target` in `_compute_target_p() `ends up having size 0, which causes problems during computation.

hanq-moreh:
```
@torch.compile(dynamic=None)
def _compute_target_p(target, t2d, loss_mask):
    target_head = target
    target_max_token = target_head.argmax(-1)
    target_mask = t2d[target_max_token]
    target_mask = target_mask[..., None].int()
    position_mask = target_mask * loss_mask
    target_head = target_head[..., t2d]
    target_head = target_head.float()
    target_p = nn.Softmax(dim=2)(target_head)
    target_p = target_p.detach()
    return target_p, position_mask
    
def _compute_target_p_padded(target, t2d, loss_mask, length):
    with torch.no_grad():
        target_p, position_mask = _compute_target_p(
            target=target,
            t2d=t2d,
            loss_mask=loss_mask,
        )

        assert len(target_p.shape) == 3
        target_p_padded = F.pad(
            target_p,
            pad=(0, 0, 0, length),
            mode="constant",
            # For bitwise equality with previous code
            value=1 / target_p.shape[-1],
        )

        return target_p_padded, position_mask
```

When cacluating target_head with t2d which is `torch.zeros` will make `target_head.shape[-1] == 0`.
And It will cause a divided by zero error in `value=1 / target_p.shape[-1]`.

specforge/utils.py
hanq-moreh: FSDP’s shard_op assumes that the optimizer is attached directly to model.parameters(), with optimizer state matching those parameters 1:1.

But class BF16Optimizer have seperate fp32 parameters currently, so as far as I tried, we can't just use FSDP's shard_op in current situation because of how BF16Optimizer is structured.

PR #484, opened 2026-03-02
fix: correct always-true condition in HarmonyParser causing duplicate system prompts

Bias92:
The condition at `parse.py` line 246 uses `or` where it should be `and` — `A != X or A != Y` is always True when X ≠ Y, so the default reasoning prompt gets prepended to every conversation regardless of the first message role. Changed to and to fix it.


jiapingW (approved):

Bias92:
thanks for approve! @jiapingW  have a good night!

PR #489, opened 2026-03-04
Fix multi-node DP training crash from FlashInfer CUDA IPC handles

yifjiang:
## Summary

- Gate CUDA IPC-based optimizations (`custom_all_reduce`, `enable_flashinfer_allreduce_fusion`) on multi-node topology detection
- Add `_is_multi_node()` and `should_disable_ipc_optimizations()` helpers with `LOCAL_WORLD_SIZE`-based detection
- Single-node training (including single-node DP) is unaffected

## Problem

When running EAGLE3 training with multi-node data parallelism (e.g. TP=4 per node, DP=2 across 2 nodes), SGLang auto-enables `enable_flashinfer_allreduce_fusion` for MoE models (`Qwen3MoeForCausalLM`) on SM90/SM100 hardware. Both this fusion and `custom_all_reduce` use `cudaIpcGetMemHandle` / `cudaIpcOpenMemHandle`, which rely on POSIX shared memory and only work between processes on the same node.

On multi-node setups this causes:
```
RuntimeError: CUDART error: invalid resource handle
```
at `flashinfer_comm_fusion.py` → `trtllm_create_ipc_workspace_for_all_reduce_fusion`.

This happens even on MNNVL-connected racks (GB200 NVL72) with IMEX channels present, because FlashInfer's IPC code uses the legacy CUDA IPC APIs, not the fabric-aware VMM APIs (`cuMemCreate` + `CU_MEM_HANDLE_TYPE_FABRIC`) that support cross-node sharing.

## Fix

- `_is_multi_node(world_size, tp_size)`: detects multi-node topology using `LOCAL_WORLD_SIZE` (set by torchrun), with fallback to `world_size > tp_size`
- `should_disable_ipc_optimizations(world_size, tp_size)`: returns `True` when multi-node is detected, `False` for single-node (including single-node DP)
- In `SGLangEagle3TargetModel.from_pretrained`: sets `disable_custom_all_reduce` and `enable_flashinfer_allreduce_fusion=False` only when `should_disable_ipc_optimizations()` returns `True`

## Test plan

- [x] 2-node GB200 NVL test (TP=4, DP=2, Qwen3-Coder-480B NVFP4): 50 steps completed successfully with IPC optimizations disabled (SLURM job 443470)
- [x] Confirmed MNNVL bypass does NOT work: keeping IPC enabled on GB200 NVL72 with IMEX channels still crashes with `CUDART error: invalid resource handle` (SLURM job 443253)
- [x] Single-node regression test (TP=4, no DP): 50 steps completed, IPC optimizations correctly kept enabled, DeviceMesh dp=1 tp=4 (SLURM job 446116)

### DIFF
--- a/specforge/modeling/target/eagle3_target_model.py
+++ b/specforge/modeling/target/eagle3_target_model.py
@@ -1,3 +1,5 @@
+import logging
+import os
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


jiapingW:
Can you help fix the conflicts?

PR #490, opened 2026-03-04
Fix eval dataset race condition in multi-node DP training

yifjiang:
## Summary

- Wrap eval dataset construction in `rank_0_priority()` so rank 0 builds/caches first
- Add `cache_dir` and `cache_key` parameters to eval dataset building for deterministic caching

## Problem

When running multi-node DP training with an eval dataset, all ranks across all nodes simultaneously spawn multiple worker processes to build the eval dataset. On NFS-backed storage, this creates race conditions where multiple processes write to the same cache files concurrently. On aarch64 (GB200 NVL), this manifests as SIGBUS crashes:

```
Bus error (core dumped)
```

For example, 16 ranks × 8 workers = 128 concurrent processes all hitting NFS at once.

## Fix

- Use the existing `rank_0_priority()` context manager (already used for training data) to let rank 0 build and cache the eval dataset first, then other ranks read from cache
- Pass `cache_dir` and `cache_key` to `build_eagle3_dataset()` for the eval dataset, matching the pattern used for training data

## Test plan

- [ ] Multi-node DP training with eval dataset runs without SIGBUS
- [ ] Eval dataset is correctly cached and reused across ranks
- [ ] Single-node training still works without regression

### DIFF
--- a/scripts/train_eagle3.py
+++ b/scripts/train_eagle3.py
@@ -485,21 +485,31 @@ def build_dataloaders(
     )
     if args.eval_data_path is not None or args.eval_hidden_states_path is not None:
         if args.eval_data_path is not None:
-            eval_dataset = Dataset.from_generator(
-                generator=safe_conversations_generator,
-                gen_kwargs={"file_path": args.eval_data_path},
+            eval_cache_params_string = (
+                f"{args.eval_data_path}-"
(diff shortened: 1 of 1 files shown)


jiapingW:
Is there any progress recently?

PR #563, opened 2026-05-26
fix: correct 9 critical/high bugs found via multi-agent audit

imitater-dou:
## Summary

This PR fixes 9 confirmed bugs discovered through a multi-agent static analysis of the codebase, all verified by direct source inspection before patching.

### 🔴 Runtime Crashes (will crash on first use)

- **`specforge/core/loss.py`** — `LogSoftmaxLoss.backward` returned 5 values for 3 inputs (`forward` takes `logits, target, position_mask`). PyTorch autograd raises `RuntimeError` on every backward pass.

- **`specforge/distributed.py`** — `Gather.backward` returned 6 values for 5 inputs (one extra trailing `None`). Crashes any distributed training run that uses sequence-parallel gather.

### 🔴 Silent Training Correctness Bugs

- **`specforge/data/preprocessing.py`** — USP `position_ids` were identical for all Ulysses ranks within the same ring group (`ring_start = ring_rank * ring_chunk` ignores `ulysses_rank`). With `sp_ulysses_size > 1`, all Ulysses peers claimed the same global position ID range, producing wrong attention patterns. Fixed by offsetting by `ulysses_rank * usp_chunk_size`.

- **`scripts/train_eagle3.py`** — `draft_model.eval()` was set at the start of the evaluation loop but never reversed. All training steps after the first eval checkpoint ran with dropout disabled. Also, `save_interval` was not scaled by `draft_accumulation_steps` (unlike `log_interval` and `eval_interval`), causing 4× more frequent saves than intended.

- **`scripts/train_dflash.py`** — On resume, only the LR scheduler state was restored (`optimizer.scheduler.load_state_dict`); the AdamW optimizer's first/second moment estimates (`exp_avg`, `exp_avg_sq`) were silently discarded, producing a loss spike at every resume. Fixed by calling `optimizer.load_state_dict(resume_state)` which restores both.

- **`specforge/lr_scheduler.py`** — `WarmupDelayerScheduler.step()` passed `epoch - warmup_epochs` to the after-scheduler, missing the `delay_epochs` subtraction. The cosine schedule started at a non-zero offset, causing a discontinuous LR jump after the warmup+delay phase.

### 🔴 Logic Bugs in Data Pipeline

- **`specforge/data/parse.py`** — `HarmonyParser` used `or` instead of `and` in the first-message guard, making the condition always `True`. Every conversation unconditionally received the reasoning effort prefix, including those that already had a system message.

- **`specforge/data/template.py`** — `TemplateRegistry.register` assert was logically inverted: `assert (not override and ...)` raised `AssertionError` whenever `override=True` was passed, making the override parameter non-functional.

### 🟠 High Severity

- **`specforge/core/eagle3.py`** — `_compute_metric_acc` used `loss_mask.sum()` as the denominator while the numerator was masked by `position_mask` (which excludes out-of-vocabulary positions). This produced systematically deflated accuracy metrics throughout training. Fixed by using `position_mask.sum()` as denominator.

- **`specforge/modeling/target/eagle3_target_model.py`** — `padding(target_out, left=False)` was called unconditionally even when `target_out is None` (the case when `return_logits=False`). This raised `AttributeError` on `None.__getitem__`. Fixed with a `None` guard.

## Test plan

- [ ] Verify `LogSoftmaxLoss` backward runs without `RuntimeError` on a minimal synthetic batch
- [ ] Verify `Gather.backward` runs without `RuntimeError` in a 2-GPU USP setup
- [ ] Verify `process_data_usp` produces distinct `position_ids` for each SP rank when `sp_ulysses_size > 1`
- [ ] Verify training resumes correctly (loss continues from pre-resume level, no spike)
- [ ] Verify `draft_model.train()` mode is active during training steps after an eval checkpoint
- [ ] Verify `TemplateRegistry.register(name, tpl, override=True)` no longer raises
- [ ] Verify `HarmonyParser` does not prepend reasoning effort prefix when the first message is a system message

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,7 +19,7 @@ jobs:
     runs-on: [self-hosted]
     container:
       image: lmsysorg/sglang:v0.5.5 # we lock to this version to avoid repeated docker pull
-      options: --gpus all --shm-size=2g --rm -v /dev/shm --privileged
+      options: --gpus all --shm-size=2g --rm -v /dev/shm --privileged --pid=host
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -47,7 +47,8 @@ jobs:
           fi
           source sf/bin/activate
           uv pip install setuptools
-          MAX_JOBS=8 uv pip install -v ".[fa]" --prerelease=allow --no-build-isolation
+          python -c "import torch" || uv pip install torch==2.9.1
+          MAX_JOBS=8 uv pip install -v ".[fa]" --no-build-isolation
 
       - name: Kill GPU processes
         shell: bash
--- a/README.md
+++ b/README.md
@@ -24,6 +24,15 @@ We have seen many open-source projects for speculative decoding, but most of the
 Check out [**our documentation**](https://docs.sglang.ai/SpecForge/) to get started.
 
 
+## 🔧 Supported Methods
+
+| Method | Description | Training | Example | Optimization |
+| --- | --- | --- | --- | --- |
+| **[EAGLE3](https://arxiv.org/abs/2503.01840)** | Feature-based autoregressive drafting | [`scripts/train_eagle3.py`](./scripts/train_eagle3.py) | [`examples/run_qwen3_8b_eagle3_online.sh`](./examples/run_qwen3_8b_eagle3_online.sh) | [LK loss](https://arxiv.org/pdf/2602.23881)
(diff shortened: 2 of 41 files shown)


jiapingW:
Hi，can you update your PR? We have fix some bugs recently.

imitater-dou:
Updated the PR with the latest upstream/main and resolved the merge conflict. It is mergeable now.

PR #581, opened 2026-06-14
feat: compact draft-vocab teacher for offline EAGLE-3 training (default off)

### LINKED ISSUE #580: [Feature] Reduce EAGLE-3 training memory with a compact (draft-vocab) teacher distribution
## Summary

In offline EAGLE-3 training, the teacher distribution is built by materializing the **full-vocabulary** target logits and casting them to fp32. In `specforge/core/eagle3.py::_compute_target_p`, `target_head = target.float()` allocates a `[B, S, vocab_size]` fp32 tensor (e.g. ~1.05 GB for `B=1, S=2048, V=128256`, and larger for Gemma-class 262K vocabularies), on top of the bf16 full-vocab logits that precede it.

However, the EAGLE-3 draft model already predicts over a **reduced draft vocabulary** (`draft_vocab_size`, via the `t2d`/`d2t` mapping). The teacher only needs the draft-vocab probabilities (`target_p`) plus a full-vocab `logsumexp` (logZ) and global `argmax` (for `target_p_on_draft` / `position_mask`, used by the acceptance / LK terms) — and these can be produced **without** ever materializing the full-vocab fp32 logit tensor.

## Proposal

Add a **config-gated, default-OFF** `--compact-teacher` path for offline training that computes the draft-vocab logits directly via the target `lm_head` **sliced to the draft-vocab rows** (`t2d`), and computes the full-vocab `logsumexp`/`argmax` with a **tiled/streaming reduction** over vocabulary chunks (O(chunk) extra memory).

This is **numerically exact-equivalent** to the current path (same `target_p`, `target_p_on_draft`, `target_token_ids`, `position_mask`, including bf16→fp32 behavior and `argmax` lowest-index tie-breaking), so the KL + acceptance/LK objective and logged metrics are preserved. With the flag off, behavior is unchanged.

## Measured benefit

On an H100, isolating the teacher-building segment at Llama3-8B dimensions (`B=1, S=2048, H=4096, V=128256, D=32000`):

| path | peak |
|------|------|
| full-vocab (current) | 3.19 GB |
| compact | 1.33 GB |
| **drop** | **1.86 GB** (≥ the eliminated fp32 `[B,S,V]` cast, 1.05 GB) |

Numerical equivalence vs the real `_compute_target_p`/`_compute_target_p_padded` verified on CPU and CUDA.

## Scope / follow-ups

This issue/PR targets the **offline** `run_forward`/`TargetHead` path (default off). Online support (avoiding `outputs.logits` allocation in the target wrappers) and opt-in experiments (draft `lm_head` init-from-verifier + freeze, an approximate/renormalized objective, a comparison harness) are deliberate follow-ups. A full end-to-end small-model training run (throughput / acceptance-length over steps) is also a follow-up; current evidence is the teacher-segment micro-benchmark plus unit/integration equivalence tests.

PR with the implementation is linked below.

Hayden727:
## Motivation

In offline EAGLE-3 training, `specforge/core/eagle3.py::_compute_target_p` builds the teacher distribution by materializing the **full-vocabulary** target logits and casting them to fp32 (`target_head = target.float()` → a `[B, S, vocab_size]` fp32 tensor, e.g. ~1.05 GB at `B=1, S=2048, V=128256`, larger for Gemma-class vocabularies), on top of the bf16 full-vocab logits.

The draft model already predicts over a reduced draft vocabulary (`t2d`/`d2t`). The teacher only needs the draft-vocab probabilities plus a full-vocab `logsumexp`/`argmax`, which can be produced **without** ever materializing the full-vocab fp32 logit tensor.

This PR adds a **default-off** `--compact-teacher` path for offline training that does exactly that, **numerically exact-equivalent** to the current path. With the flag off, training behavior and logged metrics are unchanged.

## Modifications

- `specforge/core/compact_teacher.py` (new): `compute_target_from_hidden` / `compute_target_p_padded_from_hidden` produce `(target_p, target_p_on_draft, target_token_ids, position_mask)` from the target hidden states + `lm_head` weight via a `t2d`-sliced head and a tiled fp32 `logsumexp`/`argmax` reduction (no `[B,S,vocab]` fp32 tensor). Plus validation helpers (`validate_compact_teacher_enabled`, `validate_vocab_mapping_consistency`) and the `build_offline_teacher_inputs` seam.
- `specforge/core/eagle3.py`: `OnlineEagle3Model.forward` accepts optional `target_hidden_for_compact` / `target_head_weight` / `compact_teacher_chunk_size`; when present it builds the padded teacher from hidden states and skips the full-logits path (which is otherwise untouched).
- `scripts/train_eagle3.py`: `--compact-teacher` / `--compact-teacher-chunk-size` flags (default off); offline `run_forward` delegates to `build_offline_teacher_inputs` (skips `TargetHead.forward` in compact mode); `validate_compact_teacher_args` enforces offline-only / non-VLM / valid head-shape / chunk-size / `t2d`-`d2t` consistency before training; `build_parser` / `filter_draft_state_dict` extracted as testable seams.
- Tests: `tests/test_utils/test_compact_teacher.py` and `tests/test_scripts/test_compact_teacher_integration.py`.

## Related Issues

Closes #580

## Accuracy Test

Compact-mode teacher quantities are **numerically exact-equivalent** to the existing path:

- `compute_target_from_hidden` / `compute_target_p_padded_from_hidden` match the real `_compute_target_p` / `_compute_target_p_padded` (extracted from `eagle3.py`) on **CPU and CUDA** for `target_p`, `target_p_on_draft` (within rtol=1e-4/atol=1e-5) and exact equality for `target_token_ids` / `position_mask`, including bf16→fp32 behavior and `torch.argmax` lowest-index ties across chunk boundaries.
- Acceptance-rate and KL/LK loss are equal between compact and full teacher for `lk_loss_type` `None`/`alpha`/`lambda`; a renormalized "approximate" teacher is shown to diverge when out-of-draft mass is nonzero.
- Offline `run_forward` smoke (with a `TargetHead` double) proves compact mode never calls `TargetHead.forward` while the flag-off path still does.

A full end-to-end accuracy-eval run is a follow-up; the above covers the teacher computation that this PR changes.

## Benchmark & Profiling

Isolated teacher-building segment on one H100, Llama3-8B dimensions (`B=1, S=2048, H=4096, V=128256, D=32000`), via `reset_peak_memory_stats()` + `synchronize()`:

| path | peak |
|------|------|
| full-vocab (current) | 3.186 GB |
| compact | 1.330 GB |
| **drop** | **1.856 GB** (≥ eliminated fp32 `[B,S,V]` cast = 1.051 GB) |

A patched-`F.linear` detector also asserts no compact-mode projection output has final dim == `vocab_size`. This is a teacher-segment micro-benchmark; a full end-to-end training throughput/acceptance-length comparison is a follow-up.

## Checklist

- [x] Code formatted with `black` + `isort` (pre-commit profile).
- [x] Unit + integration tests added.
- [x] Docstrings added for the new module/functions.
- [ ] Full end-to-end throughput/latency + accuracy-eval (teacher-segment micro-benchmark + numerical equivalence provided; full training run is a follow-up).

## Notes / follow-ups (out of scope here)

Online path (avoiding `outputs.logits` allocation in the HF/custom/SGLang target wrappers); opt-in experiments (draft `lm_head` init-from-verifier + freeze with an optimizer-resume guard, an approximate/renormalized objective mode, and a comparison harness/report). These can be tracked as follow-up issues.

### DIFF
--- a/scripts/train_eagle3.py
+++ b/scripts/train_eagle3.py
@@ -83,10 +83,8 @@ def print_cuda_memory_debug(label: str) -> None:
     )
 
 
-def parse_args() -> Tuple[ArgumentParser, Namespace]:
-    """
-    This function is used to parse the arguments for the training script.
-    """
+def build_parser() -> ArgumentParser:
+    """Build the training argument parser (import-safe seam for tests)."""
     parser = argparse.ArgumentParser(description="Train Eagle3 with online data")
 
     # add model-related arguments
@@ -233,6 +231,25 @@ def parse_args() -> Tuple[ArgumentParser, Namespace]:
     optimization_group.add_argument("--sp-ulysses-size", type=int, default=1)
     optimization_group.add_argument("--sp-ring-size", type=int, default=1)
     optimization_group.add_argument(
+        "--compact-teacher",
+        action="store_true",
+        help=(
+            "Offline only: compute the EAGLE-3 teacher distribution from target "
+            "hidden states via a draft-vocab-sliced head plus a streaming "
+            "logsumexp/argmax, avoiding the full-vocab [B, S, vocab] fp32 logits. "
+            "Default off; training behavior is unchanged when off."
+        ),
+    )
+    optimization_group.add_argument(
+        "--compact-teacher-chunk-size",
+        type=int,
+        default=None,
+        help=(
+            "Vocabulary chunk size for the compact-teacher streaming reduction. "
+            "Defaults to compact_teacher.DEFAULT_VOCAB_CHUNK_SIZE when unset."
+        ),
+    )
+    optimization_group.add_argument(
         "--attention-backend",
         type=str,
         default="flex_attention",
@@ -282,8 +299,40 @@ def parse_args() -> Tuple[ArgumentParser, Namespace]:
     tracker_group = parser.add_argument_group("tracker")
     TrackerArgs.add_args(tracker_group)
 
+    return parser
+
+
+def parse_args() -> Tuple[ArgumentParser, Namespace]:
+    """Parse CLI arguments for the training script."""
+    parser = build_parser()
     args = parser.parse_args()
     return parser, args
+
+
+def validate_compact_teacher_args(
+    args: Namespace,
+    is_online: bool,
+    target_model,
+    draft_model,
+    draft_model_config,
+) -> None:
+    """Validate compact-teacher settings before training (offline exact mode)."""
+    from specforge.core.compact_teacher import (
+        validate_compact_teacher_enabled,
+        validate_vocab_mapping_consistency,
+    )
+
+    target_head_weight = getattr(getattr(target_model, "fc", None), "weight", None)
+    validate_compact_teacher_enabled(
+        is_online=is_online,
+        is_vlm=args.is_vlm,
+        draft_vocab_size=draft_model_config.draft_vocab_size,
+        vocab_size=draft_model_config.vocab_size,
+        t2d=draft_model.t2d,
+        target_head_weight=target_head_weight,
+        chunk_size=args.compact_teacher_chunk_size,
+    )
+    validate_vocab_mapping_consistency(draft_model.t2d, draft_model.d2t)
 
 
 def build_tracker(args: Namespace, parser: ArgumentParser) -> Tracker:
@@ -613,6 +662,21 @@ def build_dataloaders(
     )
 
 
+def filter_draft_state_dict(model_state_dict: dict) -> dict:
+    """Keep only draft-model weights for the serving checkpoint (drop embeddings).
+
+    Embeddings are intentionally excluded (SGLang loads them from the target); the
+    target ``TargetHead`` is never part of the wrapped draft model, so no teacher state
+    can leak. The compact-teacher flag adds no new draft-model state, so its export is
+    identical to the full path.
+    """
+    return {
+        k.replace("draft_model.", ""): v
+        for k, v in model_state_dict.items()
+        if "draft_model." in k and "embed" not in k.lower()
+    }
+
+
 def save_checkpoints(
     args: Namespace,
     epoch: int,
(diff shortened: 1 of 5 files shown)


jiapingW (approved):

uygnef:
Thanks for the explanation! I agree this removes the temporary full-vocab logits spike during teacher construction.

I’m still curious about the end-to-end memory benefit. Since the full-logits path also releases the unused vocab logits after `_compute_target_p`, the later memory usage should be similar to the compact path. My understanding is that the actual peak memory is more likely to come from the later draft forward-backward stage, so this may mainly reduce the teacher-construction spike rather than the total step peak.

Also curious about the performance side: is the tiled compact path expected to be faster than the original full-logits path, or is the benefit mainly memory reduction?

Hayden727:
@uygnef Both are sharp questions:

**1. End-to-end memory**

You're right — the full-vocab `[B,S,V]` logits in the original path are transient and released right after `_compute_target_p` (the code even does `del target; torch.cuda.empty_cache()`). So what compact mode removes is mainly the *transient spike during teacher construction*, not unconditionally the total step peak.

The key point is that this transient spike still counts toward the step's `max_memory_allocated` (the caching allocator returns the block to the pool, but the high-water mark has already been recorded). So **when teacher construction happens to be the step's memory high-water mark, eliminating it does lower the total step peak** — which is more likely with a large vocabulary (e.g. Gemma 262K → ~2 GB fp32, ~3 GB including the bf16 logits), long sequences, and a small draft model (the EAGLE3 draft is typically 1 layer + fc/lm_head, whose forward/backward activations can be smaller than this full-vocab tensor). When the later draft forward-backward is the high-water mark instead, then — as you said — the benefit is mostly the spike rather than the total step peak. The 1.86 GB in the PR is exactly a **teacher-segment micro-benchmark** (it isolates just the teacher-construction segment), not the end-to-end step peak.

**2. Performance**

The benefit is **memory, not speed**. The tiled path is not faster — it still projects over the full vocabulary to compute `logZ`/`argmax` (comparable FLOPs to the original full projection), and additionally slices a separate draft-row projection for the draft probabilities (about `D/V` extra), plus the chunked-loop launch overhead. So it's **expected to be roughly on par or slightly slower** — essentially trading a bit of compute for memory. That extra draft-row projection could instead be taken from the chunk that already contains the draft rows during the tiled pass, avoiding the duplicate projection; that's a possible follow-up optimization.

FrankLeeeee:
LGTM.

PR #586, opened 2026-06-18
Expose flex_attention kernel options in DFlash and Domino training

heiheiha798:
## Summary
- add `--flex-kernel-options-json` to DFlash and Domino training CLIs
- parse the JSON dict in the training entrypoints and forward it to the online wrappers
- pass the resulting dict through to `kernel_options` in the flex attention draft forward path

## Why
On RTX A6000 (sm86), we hit a training failure when using `attention_backend=flex_attention`.

The root cause was not an NCCL error and not a generic CUDA OOM. The failure came from the Triton kernel configuration selected for flex attention during training: the generated configuration could exceed the sm86 shared-memory-per-block opt-in limit.

In practice, this means the default flex attention kernel configuration is not always viable on A6000 for these training paths.

## Local symptom on A6000
A representative workaround on RTX A6000 is to explicitly pass:

`--flex-kernel-options-json '{"num_stages": 2}'`

This reduces the selected kernel aggressiveness enough for the training path to proceed on our A6000 setup.

## Scope
This PR does not hardcode any hardware-specific default.

It only exposes the existing lower-level flex attention `kernel_options` capability through the DFlash and Domino training CLIs, so users can tune it from the command line when needed.

## Example
```bash
python scripts/train_domino.py \
  --attention-backend flex_attention \
  --flex-kernel-options-json '{"num_stages": 2}'
```

The same flag is also available in `scripts/train_dflash.py`.

## Validation
- verified that the CLI accepts a JSON dict and rejects non-dict JSON
- verified the parsed dict is forwarded to the online training wrapper
- this was added specifically to unblock A6000 flex attention training by allowing `num_stages=2` to be passed from the CLI

### DIFF
--- a/scripts/train_dflash.py
+++ b/scripts/train_dflash.py
@@ -4,6 +4,7 @@
 
 import argparse
 import functools
+import json
 import logging
 import math
 import os
@@ -66,6 +67,12 @@ def parse_args():
         default="flex_attention",
         choices=["eager", "sdpa", "flex_attention"],
         help="Attention backend for draft model.",
+    )
+    model_group.add_argument(
+        "--flex-kernel-options-json",
+        type=json.loads,
+        default=None,
+        help="JSON dict forwarded as kernel_options when attention-backend=flex_attention.",
     )
     model_group.add_argument(
         "--trust-remote-code", action="store_true", help="Trust remote code"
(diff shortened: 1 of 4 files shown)


heiheiha798:
Closing this PR and removing the source fork because it has remained untriaged without maintainer feedback for nearly four weeks, and I will no longer keep a public fork solely for this change.

This patch was intended to expose FlexAttention kernel options so that SM86 GPUs can avoid shared-memory launch failures by selecting a viable kernel configuration such as num_stages=2. If this issue remains relevant, future SM86 users will need to carry, reimplement, or upstream the fix themselves.

PR #590, opened 2026-06-23
feat: add --save-total-limit to auto-cleanup old checkpoints

jxiaof:
Training scripts save a checkpoint every --save-interval steps with no limit on how many are kept. For long runs (e.g. 500K steps with interval=2000), this produces 250+ checkpoints consuming 1.1TB+ of disk.

This PR adds a --save-total-limit argument to all three training scripts (train_domino, train_dflash, train_eagle3). When set, the oldest checkpoints beyond the limit are automatically removed after each save. Default is None (no cleanup) to preserve backward compatibility.

Changes:
- Add cleanup_checkpoints() utility in specforge/utils.py
- Add --save-total-limit arg to all three training scripts
- Call cleanup_checkpoints() after each periodic save (not after final)
- Add comprehensive unit tests in tests/test_utils/test_checkpoint_cleanup.py

## Checklist

### DIFF
--- a/specforge/utils.py
+++ b/specforge/utils.py
@@ -188,8 +188,10 @@ def print_args_with_dots(args):
 
 
 def print_on_rank0(message):
-    if dist.get_rank() == 0:
-        logger.info(message)
+    if dist.is_available() and dist.is_initialized() and dist.get_rank() != 0:
+        return
+
+    logger.info(message)
 
 
 def safe_conversations_generator(file_path):
--- a/tests/test_utils/test_print_on_rank0.py
+++ b/tests/test_utils/test_print_on_rank0.py
@@ -1,0 +1,57 @@
+import importlib
+import sys
+import unittest
+from types import ModuleType
+from unittest import mock
+
+
+def _load_utils_module():
+    transformers_module = ModuleType("transformers")
+    transformers_module.AutoTokenizer = object
+    transformers_module.PreTrainedTokenizerFast = object
+
+    transformers_utils_module = ModuleType("transformers.utils")
+    transformers_utils_module.cached_file = mock.Mock(return_value=None)
+
+    sys.modules.pop("specforge.utils", None)
+    with mock.patch.dict(
+        sys.modules,
+        {
+            "transformers": transformers_module,
+            "transformers.utils": transformers_utils_module,
+        },
+    ):
+        return importlib.import_module("specforge.utils")
+
+
+utils = _load_utils_module()
+
+
+class TestPrintOnRank0(unittest.TestCase):
+    def test_logs_without_distributed_init(self):
+        with (
+            mock.patch.object(utils, "dist") as mock_dist,
+            mock.patch.object(utils.logger, "info") as mock_info,
+        ):
+            mock_dist.is_available.return_value = False
+
+            utils.print_on_rank0("hello")
+
+        mock_info.assert_called_once_with("hello")
+
+    def test_skips_nonzero_rank(self):
+        with (
+            mock.patch.object(utils, "dist") as mock_dist,
+            mock.patch.object(utils.logger, "info") as mock_info,
+        ):
+            mock_dist.is_available.return_value = True
+            mock_dist.is_initialized.return_value = True
+            mock_dist.get_rank.return_value = 1
+
+            utils.print_on_rank0("hello")
+
+        mock_info.assert_not_called()
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_utils/test_print_on_rank0.py
+++ b/tests/test_utils/test_print_on_rank0.py
@@ -7,11 +7,11 @@ from unittest import mock
 
 def _load_utils_module():
     transformers_module = ModuleType("transformers")
-    transformers_module.AutoTokenizer = object
-    transformers_module.PreTrainedTokenizerFast = object
+    setattr(transformers_module, "AutoTokenizer", object)
+    setattr(transformers_module, "PreTrainedTokenizerFast", object)
 
     transformers_utils_module = ModuleType("transformers.utils")
-    transformers_utils_module.cached_file = mock.Mock(return_value=None)
+    setattr(transformers_utils_module, "cached_file", mock.Mock(return_value=None))
 
     sys.modules.pop("specforge.utils", None)
     with mock.patch.dict(


jxiaof:
Addressed all bot review comments and rebased the branch onto the latest main with a clean linear 2-commit history.

What changed:
- restricted checkpoint cleanup to rank 0 in distributed runs
- made print_on_rank0 safe in non-distributed contexts
- ignored FileNotFoundError during checkpoint deletion races
- gated cleanup + barrier on enabled save_total_limit only
- added focused regression tests for non-zero-rank skip and delete-race handling

Validation:
- ran python -m pre_commit run --all-files
- verified the touched files for syntax/editor errors
- ran targeted checkpoint cleanup behavior validation

Ready for human review.
@FlamingoPg @shuaills could you please take a look when you have time?

jxiaof:
@sleepcoo @shuaills

specforge/utils.py
```diff
@@ -237,6 +238,66 @@ def sort_key(x):
...
+        return (epoch, step)
+
+    checkpoints.sort(key=sort_key)
+    num_to_remove = len(checkpoints) - save_total_limit
+    removed = 0
+    for ckpt in checkpoints[:num_to_remove]:
+        ckpt_path = os.path.join(output_dir, ckpt)
+        try:
+            shutil.rmtree(ckpt_path)
+        except FileNotFoundError:
```
jessiewei7: Catching only `FileNotFoundError` means any other `OSError` from `shutil.rmtree` propagates and kills the training run. On network filesystems that's the common failure mode: e.g. NFS silly-rename leaves `.nfsXXXX` files in the checkpoint dir while another process holds an open handle, and `rmtree` fails with `ENOTEMPTY`. We hit exactly this on a live multi-node NFS-backed run.

Cleanup is best-effort — failing to delete an *old* checkpoint should never abort training:

```python
except OSError as e:
    print_on_rank0(f"Failed to remove old checkpoint {ckpt_path}: {e}; will retry on next save")
    continue
```

The leftover dir is picked up again on the next call, so nothing leaks permanently.

jessiewei7:
This wires `--save-total-limit` into `train_dflash.py` / `train_domino.py` / `train_eagle3.py`, but `scripts/train_peagle.py` has the same save path and is left out. Worth either adding it here (looks like the same three-line insertion) or noting in the description why it's excluded, so it doesn't get lost.

jxiaof:
Updates:

made checkpoint cleanup best-effort for generic OSError cases, so failures to remove old checkpoints no longer abort training
added a warning log and retry-on-next-save behavior for failed deletions
wired --save-total-limit into [train_peagle.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) as well, since it uses the same checkpoint save path
added focused regression coverage for the OSError cleanup path
Validation:

ran pre-commit on the touched files
ran focused syntax/error checks
re-validated the targeted cleanup behavior
If you have time, I’d appreciate another look.
@jessiewei7 @FlamingoPg @shuaills @sleepcoo @FrankLeeeee

specforge/utils.py
jxiaof: Thanks, this makes sense.

I updated cleanup_checkpoints so old-checkpoint deletion is now best-effort for generic OSError cases as well, with a warning and retry-on-next-save behavior instead of aborting training. I also wired the same save-total-limit handling into [train_peagle.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) so it does not get left behind.

I added focused regression coverage for the OSError cleanup path and re-ran the relevant checks.

jxiaof:
Follow-up update after the latest review comments:

- extended checkpoint cleanup to be best-effort for generic OSError cases as well, so failures to remove old checkpoints no longer abort training
- added a warning + retry-on-next-save behavior for failed deletions
- wired --save-total-limit into train_peagle.py too, since it uses the same checkpoint save path
- added focused regression coverage for the OSError cleanup path

Validation:
- ran pre-commit on the touched files
- ran focused syntax/editor checks on the touched files
- re-validated targeted checkpoint cleanup behavior, including the OSError fallback path

At this point all training scripts in this PR that save periodic checkpoints (train_domino.py, train_dflash.py, train_eagle3.py, and train_peagle.py) are covered consistently.

jiapingW:
Hi, I'm so sorry it's taking me so long to reply. Can you  update this feature to the `main`? We have updated lots of code recently. @jxiaof

jxiaof:
Hi, I rebased this branch onto the latest `main` and updated the patch accordingly.

After rebasing, I found that the original checkpoint-retention change in this PR is no longer the right integration surface on current `main`. The legacy training-script path from the original patch has since been removed/refactored, and checkpoint rotation is now handled by the unified runtime through `max_checkpoints`.

Because of that, this branch has been narrowed to the part that still applies cleanly and remains useful on top of current `main`:

- make `print_on_rank0` safe when `torch.distributed` is unavailable or not initialized
- add a focused regression test covering that behavior

I kept this update in the same PR for continuity, so the existing discussion and review history remain in one place.

Validation:
- `pytest tests/test_utils/test_print_on_rank0.py`
- `python -m pre_commit run --files specforge/utils.py tests/test_utils/test_print_on_rank0.py`

If you would still prefer this to be split into a fresh PR with the narrower scope, I can do that, but I wanted to first preserve the existing review thread and update this branch to current `main` as requested.

Please let me know if you are okay with keeping this narrowed update in PR #590.

jxiaof:
@jiapingW Thanks for the follow-up. I have now rebased this branch onto the latest `main` and pushed the updated branch.

To preserve the history of this PR, I want to explicitly summarize both the original logic in this PR and how the current codebase has changed since then.

Original logic in this PR:
- add `--save-total-limit` to the legacy training scripts
- prune older checkpoint directories after periodic saves
- keep the cleanup best-effort and safe in distributed settings
- add regression coverage around checkpoint cleanup behavior

Current logic on latest `main`:
- the legacy training-script integration path from the original patch has since been removed/refactored
- checkpoint retention/rotation is now handled by the unified runtime through `max_checkpoints`
- because of that, the original save-total-limit wiring from this PR is no longer the right surface to land on top of current `main`

So after rebasing onto current `main`, I kept the part of the original review follow-up that still applies cleanly and remains useful:
- make `print_on_rank0` safe when `torch.distributed` is unavailable or not initialized
- add a focused regression test for that behavior

I am intentionally keeping this in the same PR so the earlier discussion remains traceable and the review history is preserved, rather than losing context by starting over elsewhere.

Validation for the updated branch:
- `pytest tests/test_utils/test_print_on_rank0.py`
- `python -m pre_commit run --files specforge/utils.py tests/test_utils/test_print_on_rank0.py`

If you would like, I can also update the PR title/description to reflect the narrowed scope more accurately. If you prefer that I split this into a fresh PR instead, I can do that as well, but I wanted to first keep the existing thread intact and update the branch to current `main` as requested.

jxiaof (commented):
Addressed all bot review comments and rebased the branch onto the latest main with a clean linear 2-commit history.

What changed:
- restricted checkpoint cleanup to rank 0 in distributed runs
- made print_on_rank0 safe in non-distributed contexts
- ignored FileNotFoundError during checkpoint deletion races
- gated cleanup + barrier on enabled save_total_limit only
- added focused regression tests for non-zero-rank skip and delete-race handling

Validation:
- ran python -m pre_commit run --all-files
- verified the touched files for syntax/editor errors
- ran targeted checkpoint cleanup behavior validation

Ready for human review.
@FlamingoPg @shuaills could you please take a look when you have time?

jxiaof:
还合并吗？

PR #618, opened 2026-06-29
[DataFlow runtime] Online disaggregated training roadmap + PR plan (train-with-decode)

maocheng23:
**Docs only.** Adds `refactor/dataflow-online-roadmap.md` — the plan for taking the M6 DataFlow runtime from the shipped **offline** disaggregated path to a live **"train-with-decode"** disaggregated run (the architecture behind TorchSpec / EAGLE 3.1).

### Decision recorded
- **Regime: train-with-decode, not on-policy.** A live engine runs the *frozen target* and streams hidden states into training; the draft is not in the generation loop.
- **Consequence: weight-sync / staleness is OUT of scope** — the streamed data is independent of the draft weights, so there is nothing to re-sync (TorchSpec, the SOTA comparable, has no weight-sync either). This removes the previously-floated "weight sync + two-axis staleness gate" milestone.

### Contents
- Competitive scorecard vs **TorchSpec** (orchestration/live-gen — where we're behind) and **DeepSpec** (breadth/eval — orthogonal edge). Our data plane (`MooncakeFeatureStore`, `SampleRef`) is at parity / arguably cleaner.
- Target Ray-actor architecture.
- **PR decomposition** (the point — no mega-PRs, merge bottom-up):
  - **O1 — live single-pair:** `O1.1` shared cross-process control plane · `O1.2` `build_disagg_online_eagle3_runtime` + async loop (in-process gen stub) · `O1.3` live SGLang-server hidden-state capture 🔴 gating.
  - **O2 — Ray orchestration + scale-out:** `O2.1` actor/placement layer · `O2.2` multi-producer/trainer + DP-resharding · `O2.3` cross-pool backpressure · `O2.4` streaming-pool lifetime + online store-bug fixes.
  - Parallel **eval/breadth** track (DeepSpec parity), independent.

No code change. Opening for design review before the O1.1 PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #624, opened 2026-06-29
[DataFlow runtime · online] O1.1 — shared cross-process control plane

maocheng23:
Stage **O1.1** of the online disaggregated training roadmap (#618): a shared, durable, cross-process control plane for the online producer/consumer.

### Problem
The online producer and consumer run in separate processes, but each built its own in-process `InMemoryMetadataStore` — so commit/dedup/ack were never shared, and a restart couldn't reconcile (the producer held the committed set, the consumer held the ack marker, in two separate process-local stores). On restart the consumer also re-read the append-only channel from offset 0, silently re-training already-trained samples.

### Change
- `build_disagg_online_{producer,consumer}` accept `metadata_store` / `metadata_db_path`; both processes share one `SQLiteMetadataStore` (single-host O1 — SQLite WAL serializes the producer's commits against the consumer's ack transaction). New `_resolve_metadata_store` helper. Omitting both keeps the pre-O1.1 private in-process store.
- consumer gains `resume=`: `reconcile_on_restart` derives the already-durably-trained set and hands it to `StreamingRefQueue` as `skip_ids`, so a restarted consumer drops those refs on the channel re-read (no duplicate train); the committed-but-unacked tail re-streams and re-trains.
- `StreamingRefQueue` gains `skip_ids`: drops matching refs on read and counts them consumed so the producer's `in_flight_remote` backpressure stays exact across the restart.

### Tests
`tests/test_runtime/test_disagg_online_shared_plane.py` (CPU): cross-process commit/dedup, durable ack visible across processes, reconcile releases-acked / requeues-unacked (and requeues-all when the optimizer step wasn't durable), and restart-skip. Existing online/streaming tests unchanged.

### Scope / non-goals
SQLite is the single-host O1 tier; a `RedisMetadataStore` for multi-node O2 is a later subclass behind the same `MetadataStore` ABC. The weight-version / staleness machinery is **out of scope** per the roadmap's train-with-decode decision (#618 §1/§8) and is deliberately not in this stack.

Stacked on #622 (`dataflow-up-17-online-disagg`); O1.2 (async loop) stacks on top.

### DIFF
--- a/specforge/runtime/data_plane/streaming_ref_channel.py
+++ b/specforge/runtime/data_plane/streaming_ref_channel.py
@@ -37,7 +37,7 @@ from __future__ import annotations
 import json
 import os
 import time
-from typing import Iterator, List, Optional
+from typing import Iterator, List, Optional, Set
 
 from specforge.runtime.contracts import SampleRef, assert_no_tensors
 from specforge.runtime.data_plane.disagg_ingest import _ref_from_dict, _ref_to_dict
@@ -192,6 +192,14 @@ class StreamingRefQueue:
     the channel's consumed counter (the producer's backpressure signal); the
     feature-store free already happened in the loader's materialize (get ->
     release) for a consume-once store.
+
+    ``skip_ids`` is the restart contract (O1.1): the channel JSONL is append-only
+    and a restarted consumer re-reads it from offset 0, so without a skip set it
+    would re-train every already-durably-acked sample. The launcher derives this
+    set from the shared durable :class:`MetadataStore` (the samples
+    ``reconcile_on_restart`` reports as released) and passes it here; matching refs
+    are dropped on read and counted as consumed so the producer's backpressure
+    signal (``in_flight_remote``) stays exact across the restart.
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


maocheng23:
### Code review

No high-confidence issues found. Checked for bugs in the shared durable control-plane path, restart skip behavior, and online disaggregated launcher wiring.

maocheng23:
**Addressed review feedback (self-review pass).**

- `streaming_ref_channel.py`: `StreamingRefQueue._poll` now drains restart `skip_ids` from the set and drops to the zero-overhead fast path once the (already-trained front-prefix) skip set is empty, instead of hashing every ref for the rest of a long online run.

_Deferred (out of scope here):_ folding `StreamingRefQueue.get()` into `StreamingRefChannel.stream()` (concurrency-loop dedup) and making `reconcile_on_restart` streaming-aware (it lives in `control_plane/controller.py`, outside this PR) — both noted for follow-up.

Validated: full `tests/test_runtime` = **200 OK** (2 skipped, 1 xfail), zero failures, on a 2-node H200 pod. Lint clean (black 24.10.0 / isort 5.13.2 / autoflake).

jiapingW (approved):

PR #637, opened 2026-07-01
[DataFlow runtime] Phase D — training managers (no_sync, full resume, checkpoint/eval)

maocheng23:
# Phase D — training managers (no_sync, full resume, checkpoint/eval)

Implements **Phase D** of the reconciled plan (#630, `docs/roadmap/domain-refactor.md` §D): brings the training loop to production parity. Stacked on #636 (Phase C). Four features plus best-checkpoint tracking, each listed below with its full API surface and how it is implemented.

---

## 1. Gradient accumulation with `no_sync()`

**Where:** `specforge/runtime/training/trainer.py` (TrainerCore), `specforge/runtime/training/backend.py` (FSDPTrainingBackend).

| Function | What it does / how |
|---|---|
| `TrainerCore.train_step(batch, ctx)` | Scales loss by `accumulation_steps`, decides the optimizer boundary **before** backward (`self._micro % accumulation_steps == 0`) and passes it down, steps the optimizer only at the boundary. |
| `TrainingBackend.backward(loss, *, is_boundary=True)` (ABC) | New keyword on the seam contract. |
| `FSDPTrainingBackend.backward` | Wraps non-boundary micro-steps in `self.module.no_sync()` — the FSDP gradient reduce-scatter fires **once per optimizer step**, not once per micro-step. `accumulation_steps=1` / unwrapped modules are byte-identical to before. |

Deliberately removed: `TrainerCore.eval_step` and the `mode` plumbing — evaluation now consumes raw `forward_loss` outputs (see §4) because `_result` scalarizes away the per-position count tensors that correct acceptance-length aggregation needs.

## 2. Full resume

**Where:** `specforge/runtime/training/backend.py`, `specforge/optimizer.py`, `specforge/training/trainer.py` (domain Trainer), `specforge/runtime/data_plane/feature_dataloader.py`.

| Function | What it does / how |
|---|---|
| `FSDPTrainingBackend.state_dict()` | Returns `{model, optimizer, rng}`. `model` is gathered under `FullStateDictConfig(offload_to_cpu=True, rank0_only=True)` (no per-rank GPU replica; non-rank0 gets `{}`). `rng` holds the CPU state plus the **bound CUDA device's** state only — checkpoints are independent of `CUDA_VISIBLE_DEVICES` layout. |
| `FSDPTrainingBackend.load_state_dict(state)` | Restores model / optimizer / RNG; tolerates legacy list-valued CUDA RNG (picks the bound device's entry or fails fast with a clear message). |
| `BF16Optimizer.state_dict/load_state_dict` | Now persist the **fp32 master params** (rank-local, shape-checked on load) alongside AdamW moments and the bundled LR scheduler — resume is numerically faithful instead of re-quantizing the master from bf16 weights. Missing `fp32_params` (older checkpoint) falls back to re-clone with one warning. |
| `Trainer(resume_from=...)` (domain) | Restore order: `read_resume_state` → draft weights **before** the FSDP wrap (so the fp32 master is rebuilt from them) → optimizer/scheduler/RNG after `prepare_model` → `start_step`/`epoch`/data position. Weight load is guarded: non-empty key overlap, zero unexpected keys, and saved-`strategy`-name match, else `ValueError` (a mismatched checkpoint can no longer silently "resume" from random init). Persists and validates `dataset_size` and `accumulation_steps` — drift fails fast. |
| `FeatureDataLoader.seek(num_batches)` | Repositions the ref stream without materializing skipped features; raises when the skip exceeds the available batches. Mid-epoch position is persisted in **samples** (batch-size independent) and converted back to batches with a divisibility fail-fast. |
| `TrainerController.fit()` | Early-returns when `global_step >= max_steps` before consuming any data (idempotent restart); the mid-epoch skip is driven by the live position counter so re-entering `fit()` never re-trains the prefix; the plain-iterable skip path validates the consumed count. |

## 3. CheckpointManager

**Where:** `specforge/training/checkpoint.py` (new module).

| Function | What it does / how |
|---|---|
| `CheckpointManager(output_dir, run_id, *, max_checkpoints=0, best_metric='eval/simulated_acc_len', best_min_delta=0.0)` | One manager per run; rehydrates the best record from disk on construction. |
| `checkpoint_dir(step)` | Layout: `{output_dir}/{run_id}-step{N}/` containing `training_state.pt` (rank0 shared payload) + `training_state_rank{r}.pt` per rank — required because FSDP `use_orig_params` shards AdamW moments; rank0-only optimizer state would corrupt every other rank on resume. |
| `save(state, step, *, rank_state=None)` | All-rank `makedirs`; **atomic writes** (tmp + `os.replace`) for every payload; **timeline rewind**: rank0 deletes any on-disk step ≥ the one being saved (fork/rollback semantics, logged) and clears a best record inside the deleted range; **cross-rank failure propagation**: an `all_gather` of per-rank error strings makes one rank's FS error raise on *all* ranks instead of stranding the group in a barrier; rotation never deletes `best_step` or the just-written step, and its `rmtree` failures warn instead of killing the run. |
| `is_better(eval_metrics)` | Rank0 computes `score > best_score + best_min_delta` and the verdict is **broadcast inside the manager** — callers can't accidentally diverge rank decisions before the collective-bearing save. Empty metrics → `False`. |
| `update_best(step, eval_metrics)` | Rank0-only, atomic `{run_id}.best_meta.json` write + `{run_id}-best` repoint. Meta records `run_id` and `metric`; the loader ignores mismatched or corrupt meta with one warning. All pointer/meta artifacts are **run_id-scoped** so runs sharing an `output_dir` cannot clobber each other. |
| `read_resume_state(path, *, map_location='cpu', require_full_state=True)` | The single resume reader: merges the shared payload with this rank's file, returned untouched under `'backend'` (no key-rename seams). Fails fast on world-size mismatch and on optimizer-less checkpoints (`require_full_state=False` opts into weights+counters-only). |
| `load(step=None)` / `latest_dir()` | `latest` is trusted only when it is a real symlink pointing at a **complete** checkpoint; a materialized (copied) `latest` is repaired on the next save; the fallback scans only complete step dirs, so a truncated dir from a mid-save preemption is invisible. Relative symlinks keep pointers valid after relocating `output_dir`. |

## 4. Evaluator

**Where:** `specforge/eval/evaluator.py` (new module).

| Function | What it does / how |
|---|---|
| `Evaluator.run(model, batches, strategy, ...)` | Aggregates per-position accept **counts** across the whole eval pass and across DP ranks *before* the geometric `simulated_acc_len`, so the metric is batch-size- and rank-invariant (never a mean of ratios). Emits `eval/avg_loss`, `eval/avg_acc`, `eval/simulated_acc_len`, plus aggregated legacy `eval/acceptance_rate_i` / `eval/ploss_i` when the strategy provides them. |
| Collective schedule | Decided **globally** (one SUM over a fixed float64 slot vector, a MAX for the per-position length, one padded SUM iff any rank has per-position data): a rank with an empty or scalar-only shard issues identical collectives — hang-proof by construction. Reductions run on the rank's bound CUDA device; accumulation stays on-device (float64 counts, exact past 2^24) with one host sync. |
| Zero-batch pass | Returns `{}` — never fabricated zeros, so an empty/drained eval set cannot poison best tracking. |
| Scalar strategies | DFlash/Domino `forward_loss` now emit `accuracy_denom` (the accuracy's own denominator); the Evaluator weights scalar accuracy by it, restoring batch-size invariance for real scalar shapes (loss-token fallback documented). |
| `TrainerController.evaluate(data)` | Feeds raw `forward_loss` outputs to the Evaluator under the same `StepContext` as training (Domino's schedule-dependent loss evaluates consistently); `data=None` joins the collectives with an empty shard. Eval entry itself is a rank0-broadcast decision, and eval metrics are **logged** and merged into `last_metrics`. |

## 5. Best-checkpoint tracking

Any eval that beats the record (by more than `best_min_delta`) persists a checkpoint on demand — decoupled from `save_interval` cadence. The record survives restarts via meta rehydration, is rank-agreed by construction (§3 `is_better`), and rotation always protects it.

## Wiring

Every `build_*` entry point in `specforge/runtime/launch.py` forwards `resume_from` / `max_checkpoints`. Disagg-online builders fail fast when `resume_from` is given without `resume=True`, and when the durable ack marker is **ahead** of the checkpoint (acked-but-not-checkpointed samples would be silently skipped — the reconcile docstring now states the accurate guarantee). New managers live in their S-homes (`training/`, `eval/`); the runtime seam imports them lazily.

## Tests / gates

- `test_no_sync_equiv.py` — 2-rank FSDP, **distinct per-rank data**: the accumulation path matches per-micro-step reduction, and deferral counting pins exactly `accumulation_steps−1` deferrals per optimizer step.
- `test_checkpoint_manager.py` (new) — atomicity (truncated dirs invisible), timeline rewind, run-scoped meta/pointers, corrupt/mismatched meta, glob-metachar run_ids, materialized-`latest` repair, `is_better`/`update_best` semantics incl. `best_min_delta`, `read_resume_state` branches, and a 2-process gloo save→per-rank-read gate (dynamic port).
- `test_checkpoint_resume.py` — bit-for-bit weight round-trip with an asserted missing-key set; resume continuity vs an uninterrupted reference on distinct batches **with an explicit LR-scheduler position assert**; the production `Trainer(resume_from=...)` entrypoint end-to-end; fail-fast gates (strategy mismatch, zero-overlap, dataset-size drift, accumulation-steps drift, world-size mismatch); `max_steps` idempotent restart; re-entered `fit()` prefix safety.
- `test_evaluator_aggregation.py` — aggregate-before-geometric-sum, batch-size invariance under realistic mismatched denominators, zero-batch → `{}`, float64 counts, 2-proc gloo ragged/empty-shard schedule (scoped note: covers the Evaluator's own collectives).
- `test_feature_dataloader.py` / `test_trainer.py` — seek semantics incl. bounds + non-materialization, boundary-flag sequence on CPU, eval metrics reach the logger, checkpoint payload extras.

## Validation

- Full CPU-safe `tests/test_runtime` suite: **241 passed, 24 skipped (CUDA/multi-GPU gates), 1 xfailed** locally after the review-fix pass.
- 8×H200 pod: 239 OK (2 skip, 1 xfail) on the pre-fix revision; the GPU-gated tests (no_sync equivalence, launch equivalence) and multi-rank checkpoint paths need a pod re-run on the current head.

## Review history

Three adversarial review rounds are folded in:
1. **Review fixes 1–2** (see commit trail): per-rank resume state, data-stream seek, DP-safe evaluator schedule, durable rank-agreed best tracking, bound-device collectives.
2. **Review fixes 3** (latest): checkpoint lifecycle hardening — atomic writes, rollback/fork rewind semantics, cross-rank failure propagation, run-scoped best/latest artifacts, optimizer-less-resume fail-fast, fp32-master persistence, bound-device RNG, rank0-only gather, scalar-accuracy denominator weighting, zero-batch eval guard, eval observability, `max_steps` idempotency, embed-filter freeze check, online resume knob coupling — plus a comment-slimming pass across all touched modules and the test hardening listed above.

---

## How to review — commit by commit

The branch is structured as **5 self-contained behavioral commits**, each with its own tests and a green CPU suite at that point. Reviewing commit-by-commit is the intended path:

| # | Commit | Contract | Suite at commit |
|---|---|---|---|
| 1 | `6e1b739` | **no_sync accumulation** — boundary decided before backward; FSDP reduces once per optimizer step | 191 passed |
| 2 | `eb5e263` | **Backend full state + CheckpointManager** — `{model, optimizer(+fp32 masters), rng}`; atomic, per-rank, run-scoped layout | 210 passed |
| 3 | `e96441c` | **Full resume** — domain restore ordering, `seek`-based data repositioning, drift fail-fasts | 223 passed |
| 4 | `fa882d1` | **Evaluator** — aggregate-before-geometric-sum, DP-safe collective schedule, `accuracy_denom` weighting | 240 passed |
| 5 | `f512828` | **Best-checkpoint tracking** in the training loop; docs | 241 passed |

Rebased onto `dataflow-up-16-zerocopy` after the B/C stack (#622–#636) merged; adopts the `build_control_plane_for_mode` rename from the Phase C review fixes.


maocheng23:
One more commit from the E0/E1 review pass, placed here because it fixes this PR's code: `c71560f` — the Evaluator's scalar-accuracy path (DFlash/Domino) is now **token-weighted** (sum of correct over sum of tokens) instead of a mean of per-batch means, making it batch-size invariant like the per-position path (new gate: `test_scalar_accuracy_is_batch_size_invariant`; the DP gate's expectation updated to the weighted semantics); and `CheckpointManager`'s `best`/`latest` symlinks now use **relative** targets so `resume_from=<dir>/best` survives relocating the output directory. Revalidated: full `tests/test_runtime` = **240 OK** (2 skip, 1 xfail) on 8×H200.

jiapingW (approved):

PR #641, opened 2026-07-02
[Online track] O1.3 — live SGLang-server capture (reforward transport)

maocheng23:
## O1.3 — live SGLang-server capture (reforward transport) + spike findings

The gating spike ran on stock sglang dev (H200): per-request `return_hidden_states` exists but returns the **last layer only** — eagle3 aux capture is not reachable over the stock HTTP surface (TorchSpec closed the same gap with a 17-file sglang patch). Findings + the **upstream API proposal** (per-request aux-layer capture; optional engine-side store sink) are recorded in `docs/roadmap/o13-capture-spike.md` — as sgl-project's own framework, upstreaming is ours to land. Stacked on #638.

`SGLangServerEagle3TargetEngine` goes from stub to a working engine with the patch-free **reforward transport**: the live server does the decoding (`/generate` with raw input_ids, `--skip-tokenizer-init`, greedy default → reproducible frozen-target stream); an in-process capture engine over the same weights (hf/sglang/custom — the extraction-gate-validated backends) runs one extend over `[prompt+completion]` for aux+target features. Same `Eagle3TargetOutput` contract as every backend — RolloutWorker/FeatureStore/trainer see nothing new; an upstreamed engine-side transport later replaces the reforward without touching callers. Loss is masked to the generated region, with the final position zeroed (left-shifted-target convention).

**Gate** (`test_o13_server_capture.py`, opt-in `SPECFORGE_RUN_SERVER_TESTS=1`): boots a real SGLang server, live decode → reforward capture → **one real train step with zero precomputed features**; frozen-target reproducibility (same request twice → same tokens, bit-identical features); loss region = generated tokens minus the teacherless final position. **Validated on H200: suite 241 OK + server gate OK.** An adversarial review confirmed 4 defects (teacher-alignment on the final loss position — high; capture-contract surface; timeout routing; kwargs routing) — all fixed here.

### DIFF
--- a/docs/roadmap/o13-capture-spike.md
+++ b/docs/roadmap/o13-capture-spike.md
@@ -1,0 +1,58 @@
+# O1.3 Capture Spike — findings & transport decision (July 2026)
+
+Run on: `lmsysorg/sglang` dev image (sglang `0.0.0.dev1+g70df09b83`, torch 2.11,
+H200), stock server, tiny + real-model probes over HTTP.
+
+## What the stock SGLang server exposes today
+
+| Surface | Verdict |
+|---|---|
+| `GenerateReqInput.return_hidden_states` (+ `--enable-return-hidden-states`) | **Yes** — per request, `meta_info["hidden_states"]`: one entry per forward pass (the prefill entry carries every prompt position), **LAST layer only** |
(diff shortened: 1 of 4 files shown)


jiapingW:
LGTM.

PR #653, opened 2026-07-06
[Online track] O2 — server-side spec capture: sglang 0.5.14 patch + zero-copy Mooncake transport

maocheng23:
# [Online track] O2 — server-side spec capture: sglang 0.5.14 patch + zero-copy Mooncake transport into the data plane

## What this is

The **engine-side transport** that PR #641's reforward engine anticipated ("a future engine-side transport replaces step 2 without touching callers"). A live SGLang server — patched with a versioned, self-contained patch — captures aux/last hidden states **during its own prefill forward** and writes them **straight into the Mooncake store** in `MooncakeFeatureStore`'s zero-copy key layout. The rollout side builds committed-ready `SampleRef`s from response *metadata alone*; the trainer's existing `MooncakeFeatureStore.get()` reads the server-written bytes with **zero changes**. Feature tensors never cross the HTTP response, the rollout worker, or any re-serialization — one forward per sample, engine-side capture, RDMA-capable transport.

```
      producer pool                                     consumer pool
 ┌──────────────────────┐        ┌────────────┐        ┌─────────────────────┐
 │ patched SGLang server │ put() │  Mooncake  │ get()  │ FeatureDataLoader   │
 │ (--enable-spec-capture)──────▶│   store    │◀───────│  -> TrainStrategy   │
 └──────────▲───────────┘  RDMA/ └────────────┘  zero- └─────────────────────┘
            │ /generate + spec_capture         copy
 ┌──────────┴───────────┐  keys/shapes only   ┌─────────────────────┐
 │ SGLangServerCapture-  │────────────────────▶│ RolloutWorker       │
 │ Adapter (RefSource)   │      SampleRef      │ commit_samples      │
 └──────────────────────┘                      └─────────────────────┘
```

## vs the alternatives

| | in-process (`sglang` backend) | #641 reforward | **this PR** |
|---|---|---|---|
| forwards per sample | 1 | 2 (decode + re-capture) | **1** |
| sglang fork surface | 4 files (white-box ModelRunner) | 0 | versioned patch, 8 small hooks + 1 self-contained module |
| tensors travel | in-process | in-process (capture half) | **server → Mooncake → trainer, zero-copy** |
| inference/training scaling | colocated | server decouples decode only | **fully disaggregated** |

TorchSpec closed the same gap with a 16-file engine fork; this patch is deliberately smaller because sglang 0.5.14 already ships most of the machinery (`CaptureHiddenMode.FULL` returns the aux concatenation; `ModelRunner.init_aux_hidden_state_capture()` exists with a literal `TODO: expose this to server args?` — the patch answers that TODO, which is also the upstream API proposal from the O1.3 spike).

## The sglang patch (`patches/sglang/v0.5.14/spec-capture.patch`)

410 insertions: 8 existing files get thin hooks (~150 lines), all capture/store logic lives in ONE new self-contained module (`sglang/srt/spec_capture_sink.py`):

- `--enable-spec-capture` + `--spec-capture-aux-layer-ids`: aux capture on the target **without a speculative draft worker** (requires `--chunked-prefill-size -1`, asserted at boot)
- per-request `spec_capture` dict threaded `GenerateReqInput → Req` (one optional field; rides the existing `return_hidden_states`/`CaptureHiddenMode.FULL` path)
- `LogitsProcessorOutput.last_hidden_states`: under FULL+aux, the post-norm last hidden is exposed alongside the aux concatenation (the `target_repr="hidden_state"` representation)
- scheduler sink: per-request tensor slices → `put_from` raw bytes at `{store_id}/{sample_id}/g{gen}/{name}`, hard-pinned (SpecForge, not Mooncake LRU, is the lifetime authority); results ride the existing `customized_info` channel into `meta_info["spec_capture"]` — **no new IPC fields**
- failure → per-request `meta_info["spec_capture_error"]` + best-effort key cleanup (a partial sample is never consumable)

Apply to an installed sglang: `scripts/apply_sglang_spec_capture_patch.sh`.

## Strategy-agnostic (eagle3 + dflash + domino, and future drafts)

The server knows only two generic artifacts (`aux`, `last_hidden`) plus verbatim passthrough tensors. Feature **naming and shapes are decided per request by the client** via `ServerCaptureSchema` (`specforge/inference/adapters/server_capture.py`):

- **eagle3**: `aux → hidden_state`, `last_hidden → target` (`target_repr="hidden_state"`: the trainer's frozen `TargetHead` recomputes logits — storing (L, vocab) logits would be ~50× the traffic for no benefit), `loss_mask` stored `(1, L, 1)`
- **dflash / domino**: `aux → hidden_states`, no target, `loss_mask` `(1, L)`
- new strategies: `register_server_capture_schema(...)`

## Data-plane integration

- `RolloutWorker` gains the `RefSource` path: `produce_refs → verify-from-specs → commit`; per-task failure markers keep the lease bookkeeping exact (every leased task still ends in exactly one terminal controller action)
- `verify_feature_contract_specs()`: the same loud extraction-boundary checks (names / aux-layer ids / aux width / target dim), run from the returned `FeatureSpec` shapes — no tensor fetch
- `MooncakeFeatureStore.adopt(ref)`: registers server-written samples so producer-side `abort()`/`gc()` frees them; the consumer side needed **zero changes** (its consumer-cache path already handles refs it never put)

## Tests

- `tests/test_runtime/test_server_capture.py` (CPU, in the default suite): a stub server emulates the sink byte-for-byte against the fake Mooncake backend; drives the real adapter → worker → store stack. Covers: eagle3 zero-copy roundtrip (bit-exact), dflash naming, aux-width mismatch fails loud **and frees keys**, per-task server errors, worker commit/fail bookkeeping, adopt→abort, retry generation bump, spec-based verification parity.
- `tests/test_runtime/test_server_capture_gate.py` (opt-in `SPECFORGE_RUN_SERVER_CAPTURE_TESTS=1`, GPU): boots a **real patched server + real mooncake master**; eagle3 refs → zero-copy `get()` → aux parity vs an independent HF `output_hidden_states` forward + `lm_head(last_hidden)` vs HF logits (bf16 `2e-2`, the extraction-gate tolerance) → **one real train step** with `target_repr="hidden_state"`; dflash capture on the same server.

## Validation (sci-h200, 8×H200, sglang==0.5.14 pip + patch applied, torch 2.11+cu130, live `mooncake_master`, TCP)

- patch applies cleanly to pristine `v0.5.14` (`git apply --check`) and to an installed sglang via the helper script (`patch -p2`)
- **GPU gate** (`SPECFORGE_RUN_SERVER_CAPTURE_TESTS=1`): **2/2 OK** — eagle3: refs → zero-copy `get()` → aux parity vs HF `output_hidden_states` + `lm_head(last_hidden)` vs HF logits (bf16 2e-2) → one finite train step; dflash capture on the same server
- **full `tests/test_runtime` regression: 310 OK (skipped=4, xfail=1), zero failures** (includes the 9 new CPU contract tests)
- the gate caught a real integration bug during bring-up (eagle3 `loss_mask` must follow the offline `(B, L)` convention — `TargetHead.preprocess` adds the mask dim); fixed in this PR

## Relationship to PR #641

#641 (reforward, patch-free, base `dataflow-mla-draft`) stays valuable as the zero-patch fallback and for live-decode online capture; this PR is the transport its engine docstring anticipated. The two compose: #641's server decode + this patch's capture-on-prefill = one-forward online capture with generation (follow-up). Long-term both converge on upstreaming the capture API into sglang (`docs/roadmap/o13-capture-spike.md`) — this patch's server-args surface is that proposal, validated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,9 @@ repos:
       - id: check-symlinks
       - id: destroyed-symlinks
       - id: trailing-whitespace
+        exclude: \.patch$  # stripping context-line whitespace breaks git apply
       - id: end-of-file-fixer
+        exclude: \.patch$
       - id: check-yaml
         args: [--allow-multiple-documents]
       - id: check-toml
--- a/examples/disagg/README.md
+++ b/examples/disagg/README.md
@@ -1,3 +1,14 @@
+# Disaggregated examples
+
+Two flavors live here:
+
+- **Online DFlash via server-capture** (`run_disagg_dflash.py` +
+  `run_qwen3.6_27b_dflash_disagg.sh`) — the real zero-copy split: a live patched
+  SGLang server captures features and writes them straight into Mooncake; the
+  trainer consumes them by key. See [Online DFlash](#online-dflash-via-server-capture-real-zero-copy).
+- **Offline EAGLE3** (`run_disagg_eagle3.py`) — precomputed features ingested
+  into a shared store, documented below.
+
 # Disaggregated offline EAGLE3 example
 
 Runs the offline EAGLE3 training of `scripts/train_eagle3_dataflow.py`, but splits
@@ -99,3 +110,46 @@ acc / acceptance_rate climb over training in both (baseline direction). Per-step
 values are noisy at `batch_size=1` over 64 diverse samples. Note this is the
 training-time acceptance proxy; the serving accept-length (τ via spec-decoding) is
 a separate eval gate.
+
+# Online DFlash via server-capture (real zero-copy)
+
+The engine-side transport: a live SGLang server does the capture and writes
+straight into Mooncake, so inference and training are fully separate processes
+that share only the object store — no in-process target, no shared *data* mount.
+
+```
+ inference pool (GPU)                 mooncake                training pool (GPU)
+ ────────────────────                ─────────               ───────────────────
+ patched SGLang server  ──put()────▶  object    ──get()────▶ FeatureDataLoader
+ (--enable-spec-capture)   RDMA/TCP    store       zero-copy  -> OnlineDFlashModel
+        ▲                                                     TrainerController.fit()
+        │ /generate + spec_capture   (keys only)
+ SGLangServerCaptureAdapter ───────▶ StreamingRefChannel ───▶ (consumer reads refs)
+```
+
+The server is stock `sglang==0.5.14` patched with
+`patches/sglang/v0.5.14/spec-capture.patch` — apply it to the installed package
+with `scripts/apply_sglang_spec_capture_patch.sh`. Only tiny `SampleRef`s cross
+the ref channel; the DFlash context hidden states go server → Mooncake →
+trainer with no re-copy. Strategy-agnostic: the same server serves eagle3 or
+dflash requests, named per request by the client
+(`specforge/inference/adapters/server_capture.py`).
+
+## Run it (single 8-GPU node)
+
+```bash
+export WANDB_API_KEY=<key>            # or add --report-to none in the wrapper
+bash examples/disagg/run_qwen3.6_27b_dflash_disagg.sh
+```
+
+The wrapper starts a `mooncake_master`, the patched server (GPU 0), a thin CPU
+producer, and the trainer (GPU 1). `--spec-capture-aux-layer-ids` must match
+`dflash_config.target_layer_ids` in `configs/qwen3.6-27b-dflash.json` (the
+producer reads the same list for contract verification). Mooncake connection is
+the standard `MOONCAKE_*` env (see the wrapper for defaults).
+
+The transport is gate-validated end-to-end (patched server → Mooncake → one real
+train step, with aux parity vs an HF reference) in
+`tests/test_runtime/test_server_capture_gate.py`
+(`SPECFORGE_RUN_SERVER_CAPTURE_TESTS=1`, GPU); the contract/ref/adapter logic is
+covered on CPU in `tests/test_runtime/test_server_capture.py`.
--- a/examples/disagg/run_disagg_dflash.py
+++ b/examples/disagg/run_disagg_dflash.py
@@ -1,44 +1,40 @@
-"""Disaggregated ONLINE DFlash example: rollout/inference pool and trainer pool
-as separate processes that share only a filesystem mount.
+"""Disaggregated ONLINE DFlash example — real server-capture zero-copy transport.
 
-Online analogue of ``run_disagg_eagle3.py`` (which wires the *offline* seam). Here
-the target model is never precomputed to disk — the **producer** runs it live and
-streams captured features to the **consumer**, which trains the DFlash draft:
+Inference and training are fully decoupled processes that share only a Mooncake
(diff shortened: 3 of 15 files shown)


maocheng23:
Follow-up pushed (`f60a6a1`, `9b0ebce`):

- **The disaggregated DFlash-Qwen36 example is now the real server-capture version.** `examples/disagg/run_disagg_dflash.py` producer is a thin HTTP driver (no in-process target); `run_qwen3.6_27b_dflash_disagg.sh` launches `mooncake_master` + the patched server (`--enable-spec-capture --spec-capture-aux-layer-ids ...`) + producer + trainer. README section added.
- `launch.py`: `build_disagg_online_producer` / `_assemble_rollout_workers` accept an optional `feature_source` (RefSource) so the server-capture adapter drives the producer without an in-process `target_model`.
- Trimmed the over-verbose comments/docstrings across the patch + new modules; regenerated the patch (still 9 files, clean apply on pristine v0.5.14).
- Excluded `*.patch` from the whitespace/eof pre-commit hooks (they strip context-line whitespace and break `git apply`).
- Added a CPU test for the `feature_source` producer wiring.

Re-validated on sci-h200: **311 test_runtime OK**, GPU gate **2/2 OK**, pre-commit clean.

jiapingW (approved):

maocheng23:
**End-to-end 5000-step disaggregated run: DONE** (Qwen3.6-27B DFlash, single node: patched server TP=4 on GPUs 0-3, thin producer on GPU 5, single-DP trainer on GPU 4, live `mooncake_master`, TCP).

Windowed training curve (500-step bucket means, batch=1):

| steps | loss | acc |
|---|---|---|
| 0-499 | 6.66 | 0.052 |
| 1000-1499 | 5.54 | 0.106 |
| 2500-2999 | 4.80 | 0.149 |
| 4000-4499 | 4.98 | 0.154 |
| 4500-4999 | **4.59** | **0.178** |

33 min wall (~2.6 steps/s, producer pinned at the 256-sample watermark ahead of the trainer the whole run), 5256 samples captured server-side, **zero sink put failures, zero eviction triggers**.

Two production bugs found by the run and fixed here:

1. `mooncake: never probe existence before a remove` — probed on a live master: `is_exist` **grants a read lease**, so a remove right after it fails (-706). `gc()`'s exists pre-check therefore re-leased every pending object right before its retry, exhausted `max_release_attempts`, and leaked 100% of consumed samples (64GB and 200GB segments both filled at exactly cumulative-produced bytes). Removes now run first; the exist probe only classifies already-gone keys after a failed remove.
2. `data-plane: loader pumps store.gc() on a lease-aware cadence` — releases are lease-deferred by design, but nothing in the consumer loop retried the parked frees. `FeatureDataLoader` now pumps `gc()` every `gc_interval_s` (default 15s > lease TTL).

Plus the disagg example config hardening: producer/consumer are **pure Mooncake clients** (`global_segment_size=0`) so no committed feature can live in a process that exits early; the server sink's segment is the only storage, sized above the in-flight watermark.

Re-validated: full `test_runtime` **312 OK** (incl. a new lease-semantics regression test for the gc ordering), GPU gate OK.

PR #659, opened 2026-07-07
fix(train_dflash): atomic per-file checkpoint writes + `_SUCCESS`-gated resume

jessiewei7:
A runtime checkpoint is now consumable only after its `_SUCCESS` marker is published. Resume, export, runtime warm start, online checkpoint reconciliation, and training gates share that rule, so an interrupted save cannot be selected or consumed through a side path.

`_SUCCESS` is published only after every rank has written its payload, the write is confirmed collectively, and rank 0's publish is itself confirmed — before the `*-latest` pointer moves. Payloads, the marker, and the parent directory are fsynced so a crash cannot leave a marked-but-partial checkpoint.

### DIFF
--- a/specforge/export/checkpoint_io.py
+++ b/specforge/export/checkpoint_io.py
@@ -10,64 +10,27 @@
 
 from __future__ import annotations
 
-import glob
 import os
-import re
 from typing import Any, Dict, Optional
 
 import torch
 
-STATE_FILE = "training_state.pt"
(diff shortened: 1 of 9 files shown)


jiapingW:
Hi, can you update your code to current `main` branch?

jessiewei7:
Yes, I've updated the PR for the current main branch.

PR #242, opened 2025-09-26
Apply FSDP2 to offline training

j1young:
## Motivation
apply FSDP2 to the offline training pipeline.

On H100x4 GPUs, the training throughput shows an improvement of ~1.4× compared to the previous baseline.

The following command was used for testing
```
torchrun \
    --standalone \
    --nproc_per_node 4 \
    ./scripts/train_eagle3_offline.py \
    --target-model-path lmsys/gpt-oss-120b-bf16 \
    --draft-model-config ./configs/gpt-oss-120B-eagle3.json \
    --train-data-path ${TRAIN_DATA_PATH} \
    --train-hidden-states-path ${TRAIN_HS_PATH} \
    --output-dir ${OUTPUT_DIR} \
    --draft-global-batch-size 16 \
    --draft-micro-batch-size 1 \
    --num-epochs 10 \
    --learning-rate 1e-4 \
    --max-length 2048 \
    --chat-template gpt-oss
```

## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://sgl-fru7574.slack.com/archives/C09784E3EN6 to discuss your PR.

PR #660, opened 2026-07-07
[DataFlow runtime] Domino disaggregated training (example + launcher + strategy params + capture/tracker fixes)

maocheng23:
## Summary

Adds end-to-end **Domino disaggregated training** on top of the DataFlow runtime integration branch (`dataflow-up-16-zerocopy`), plus a few robustness fixes surfaced along the way. Cleanly stacked: **2 commits, +718 / −71 across 12 files**, no rebase needed.

## Commits

| SHA | Author | Message |
|-----|--------|---------|
| `6cd6829` | canghua | support domino disaggregated training[bug] |
| `7579d8e` | wjp666666 | fix bugs |

## What's in it

**New Domino disagg entrypoints**
- `examples/disagg/run_disagg_domino.py` (+298) — end-to-end domino disaggregated training example
- `examples/.../run_qwen3_8b_domino_disagg_multiserver.sh` (+216) — multiserver launcher; plus edits to the qwen3.6-27b dflash disagg launcher
- `configs/qwen3-8b-domino.json` — adds `emb_dim: 256`

**Parameterized domino strategy**
- `training/strategies/registry.py` — domino strategy now accepts `lambda_start` / `decay_ratio` via a new `_make_domino_strategy` builder
- `training/trainer.py` + `launch.py` — thread a new `strategy_kwargs` dict through `Trainer`, `_assemble_trainer`, and `build_disagg_online_consumer` so strategy construction can be parameterized
- `registry.py` — **online collate fix**: both `dflash` and `domino` now use `_dflash_offline_collate` for the online path instead of `concat_collate` (behavioral change — worth a close look)

**Robustness fixes**
- `inference/adapters/server_capture.py` (+51) — handle batch/list-wrapped capture results from the spec-capture server; select this task's row by `sample_id`, with explicit errors on ambiguous or non-object results (`_flatten_list_wrappers`, `_capture_result_for_task`)
- `tracker.py` — treat a local `wandb/` output dir mis-imported as a namespace package as "wandb unavailable"; raise a clear error if `--report-to wandb` is used without the real client
- `runtime/data_plane/mooncake_store.py` — include the underlying exception + CUDA-runtime hint in the mooncake load-failure message

**Tests**
- `tests/test_runtime/test_server_capture.py` (+19) — batch-result selection
- `tests/test_runtime/test_strategy_registry.py` (+22) — domino strategy params / online collate

## Reviewer notes

- Commit messages are terse (`fix bugs`, `[bug]`) — worth squashing/rewording before merge.
[...]

PR #663, opened 2026-07-08
fix(optimizer): clip by global grad norm across FSDP ranks in BF16Optimizer

jessiewei7:
# fix(optimizer): clip by global grad norm across FSDP ranks in BF16Optimizer

## Summary

`BF16Optimizer.step()` clips with plain `torch.nn.utils.clip_grad_norm_` over
its fp32 master copies. All callers (`train_dflash.py`, `train_domino.py`,
`train_eagle3.py`, `train_peagle.py`, the dataflow runtime, the disagg
example) wrap the model in FSDP with `SHARD_GRAD_OP`, so after backward each
rank's `.grad`s are its reduce-scattered **shard** — the clip sees a
rank-local norm, roughly `1/sqrt(world_size)` of the true norm. Two
consequences:

- gradients are systematically under-clipped (the threshold effectively
  scales with world size), and
- each rank picks a **different** clip coefficient for its shard, so the
  clipped global gradient no longer points in the original direction.

`specforge/runtime/training/backend.py` already acknowledges this — its
`step()` all-reduces the norm — but it does so **after** `optimizer.step()`
has clipped, so only the *logged* norm was corrected, not the clipping
itself.

## Fix

- `BF16Optimizer._clip_grad_norm()`: accumulate the squared norm, all-reduce
  it across ranks (when `dist` is initialized), and scale every shard by the
  single global clip coefficient. Non-distributed behavior is unchanged
  (verified against `torch.nn.utils.clip_grad_norm_` in the test). The
  returned/logged norm is now the true global norm.
- `FSDPTrainingBackend.step()`: drop the post-hoc all-reduce — with the fix
  it would reduce twice and inflate the logged norm by ~`sqrt(world_size)`.

Why not `FSDP.clip_grad_norm_`: it clips the module's bf16 shard gradients,
while `BF16Optimizer`'s whole design is clipping its fp32 master copies —
this keeps that numerics path, applying the same cross-rank reduction that
API does internally. An FSDP2 migration would make this class of bug
impossible (DTensor grads reduce cross-rank automatically) — #242 did that
for the old offline script, but the current dataflow backend is back on
FSDP1 — so the FSDP1 paths need the reduction done explicitly, as here.

Note on behavior: `max_grad_norm=0.5` was copied from EAGLE's DeepSpeed
`ds_config.json`, and DeepSpeed clips by the global norm — so this restores
parity with the reference implementation the constant came from. At scale
the true norm is ~`sqrt(world_size)`× the previously-compared local norm, so
clipping engages more often; existing multi-GPU loss curves will not
reproduce exactly.

## Test plan

`tests/test_optimizer/test_bf16_optimizer_clip_grad_norm.py` (CUDA-free):

- single-process clipping matches `torch.nn.utils.clip_grad_norm_` exactly
  (norm and clipped gradients);
- 2-process `gloo` run with different per-rank gradient shards: both ranks
  compute the same global norm and apply the same clip coefficient.

### DIFF
--- a/specforge/optimizer.py
+++ b/specforge/optimizer.py
@@ -1,4 +1,5 @@
 import torch
+import torch.distributed as dist
 
 from specforge.lr_scheduler import CosineAnnealingWarmupLR
 from specforge.utils import print_on_rank0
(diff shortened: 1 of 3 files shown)


jiapingW:
Great findings! We have fixed the bugs in [branch](https://github.com/sgl-project/SpecForge/pull/660/changes/6cd6829ab20ef5a94d0020414e9ec21249638026).

jiapingW:
Hi. We have fixed this bug in `main` branch [here](https://github.com/sgl-project/SpecForge/blob/main/specforge/optimizer.py#L56). You can check if it aligns with your expected implementation. Thank you very much for your contribution. Please close this PR if you think this issue is resolved.

jessiewei7:
Confirmed — the main implementation addresses this issue, including the NO_SHARD case. I've closed the PR. Thanks for the update.

Issue #466, opened 2026-02-12
[Bug] Abnormal memory usage and Out-of-Memory in eagle3 training.


coco-alen:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/SpecForge/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When training an Eagle model for qwen3, I could only set the batch size to 2; increasing it further caused an OutOfMemoryError (OOM). I was using 4 H200 GPUs, each with 140GB of VRAM. This VRAM usage was clearly abnormal.

I found that most of the OOMs occurred in the `padding` function within `specforge/utils.py`.

### Reproduction

```python
NUM_GPUS=${1:-4}

#torch memory segment
export PYTORCH_ALLOC_CONF=expandable_segments:True
CUDA_VISIBLE_DEVICES=4,5,6,7 \
torchrun \
    --standalone \
    --nproc_per_node $NUM_GPUS \
    $ROOT_DIR/scripts/train_eagle3.py \
    --target-model-path Qwen/Qwen3-8B \
    --draft-model-config $ROOT_DIR/configs/qwen3-8b-eagle3.json \
    --train-data-path $ROOT_DIR/cache/dataset/dflash_dataset_nemotron_800K.jsonl \
    --output-dir $ROOT_DIR/outputs/qwen3-8b-eagle-nemotron \
    --num-epochs 10 \
    --batch-size 2 \
    --learning-rate 2e-4 \
    --max-length 4096 \
    --chat-template qwen \
    --cache-dir $ROOT_DIR/cache \
    --embedding-key model.embed_tokens.weight \
    --log-interval 50 \
    --save-interval 1000 \
    --tp-size $NUM_GPUS \
    --target-model-backend sglang \
    --sglang-attention-backend fa3 \
    --sglang-mem-fraction-static 0.1 \
    --report-to wandb \
    --wandb-project specforge-qwen3-8b-eagle \
    --wandb-name qwen3-8b-eagle-nemotron
```

### Environment

specforge                 0.2.0

[...]

PR #669, opened 2026-07-12
feat(training): gradient checkpointing for the TTT unroll (`--draft-gradient-checkpointing`)

notwitcheer:
## Motivation

The TTT loop unrolls `ttt_length` draft forwards and retains every step's activations until backward: attention probs, MLP intermediates, the per-step `seq x draft_vocab` logits (131 MB per step at max-length 2048 with a 32K draft vocab), and the per-step teacher-p slices (~262 MB fp32 per step). At longer max-length this retained graph, not the weights, is what OOMs online training on a single card. Related memory asks: #466, #528, #558 (#581 addressed the teacher side, offline only).

## What this does

`--draft-gradient-checkpointing` (default off, no behavior change when off) recomputes each TTT step during backward instead of retaining it. Naive `torch.utils.checkpoint` around a step is incorrect here, not a speed tradeoff: backward replays the forward, and a replay that appends to a shared KV cache appends a second time, corrupting every later step. The step function is made replay-safe:

- Prior steps' cached k/v enter the checkpointed function as explicit tensor arguments, so the cross-step gradient chain survives recompute.
- A fresh local cache is built per invocation (`cache_hidden` lists for sdpa/fa; a `DynamicCache` seeded via `.update()` for flex_attention, with the transformers>=5 `cache.layers[i]` API and legacy fallback), so a replay can never double-append.
- The loss is computed inside the checkpointed region, so the per-step logits and teacher-p slices stay recompute-transient instead of being retained by the loss graph.

`usp` is rejected with a `ValueError` (it runs collectives inside the step, which is not replay-safe). The QwenVL online path rejects the flag rather than silently ignoring it. The online and offline text paths both go through `OnlineEagle3Model` and both support it.

## Verification

The equivalence bar is bitwise, not a tolerance:

1. Run the unpatched path twice on identical inputs to measure the backward noise floor: 0.0.
2. Require checkpointed-vs-unpatched gradient diff <= that floor.

Result on a real 1-layer draft head (155K target vocab / 32K draft vocab, flex_attention): all 7 step losses and every parameter gradient bitwise identical (diff exactly 0.0). Anything above the floor would mean the replay is not reproducing the forward.

Production evidence: a 7,000-step online training night at max-length 2048 on one RTX 5090 (36B w4a16 teacher via the sglang backend), rc=0, ~3.7 s/step vs ~3.5 s/step at 1024: the extra draft forward per step drowns under a teacher-bound step. Retained graph dropped ~85%; 2048 previously missed single-card fit by ~250 MB with every other memory option enabled, and now fits with headroom. A spec-decode eval of the resulting checkpoint served normally in sglang EAGLE-3.

## Checklist notes

- pre-commit clean.
- I have the equivalence A/B as a standalone script and can adapt it into a GPU unit test in this PR if you want it in CI.

### DIFF
--- a/scripts/train_eagle3.py
+++ b/scripts/train_eagle3.py
@@ -261,6 +261,18 @@ def build_parser() -> ArgumentParser:
         default="flex_attention",
         help="The attention backend for the draft model",
     )
+    optimization_group.add_argument(
+        "--draft-gradient-checkpointing",
+        action="store_true",
+        help=(
+            "Recompute each TTT step during backward instead of retaining its "
+            "activations. Cuts the retained graph from ~O(ttt_length * "
+            "seq * vocab/intermediate) to just the per-step hidden states and "
+            "cached k/v, at the cost of one extra draft forward per step. "
+            "Enables longer --max-length on a single card. Not supported with "
+            "the usp attention backend."
+        ),
+    )
 
     # other args
     other_group = parser.add_argument_group("others")
@@ -1064,6 +1076,11 @@ def main():
         and args.tp_size == 1
         and args.target_model_backend != "sglang"
     ):
+        if args.draft_gradient_checkpointing:
+            raise ValueError(
+                "--draft-gradient-checkpointing is not supported for the "
+                "QwenVL online path"
+            )
         eagle3_model = QwenVLOnlineEagle3Model(
             target_model=target_model,
             draft_model=draft_model,
@@ -1084,6 +1101,7 @@ def main():
                 lk_loss_type=args.lk_loss_type,
                 kl_scale=args.kl_scale,
                 kl_decay=args.kl_decay,
+                gradient_checkpointing=args.draft_gradient_checkpointing,
             )
         else:
             # offline: the target_model is TargetHead not a model
@@ -1094,6 +1112,7 @@ def main():
                 lk_loss_type=args.lk_loss_type,
                 kl_scale=args.kl_scale,
                 kl_decay=args.kl_decay,
+                gradient_checkpointing=args.draft_gradient_checkpointing,
             )
     eagle3_model = FSDP(
         eagle3_model,
--- a/specforge/core/eagle3.py
+++ b/specforge/core/eagle3.py
@@ -25,6 +25,7 @@ from typing import Callable, List, Optional, Tuple
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+import torch.utils.checkpoint
 from transformers.cache_utils import DynamicCache
 
 from specforge.core.compact_teacher import (
@@ -40,6 +41,13 @@ from specforge.utils import padding
 
 class Eagle3Model(nn.Module):
     pass
+
+
+def _dynamic_cache_kv(cache: DynamicCache) -> Tuple[torch.Tensor, torch.Tensor]:
+    """Return the layer-0 key/value tensors across transformers versions."""
+    if hasattr(cache, "layers"):  # transformers >= 5
+        return cache.layers[0].keys, cache.layers[0].values
+    return cache.key_cache[0], cache.value_cache[0]
 
 
 def _compute_loss_and_acceptance_rate(
@@ -116,6 +124,7 @@ class OnlineEagle3Model(Eagle3Model):
         lk_loss_type: Optional[str] = None,
         kl_scale: float = 1.0,
         kl_decay: float = 1.0,
+        gradient_checkpointing: bool = False,
     ):
         """
         Args:
@@ -125,6 +134,12 @@ class OnlineEagle3Model(Eagle3Model):
             lk_loss_type: LK loss objective type. One of {"lambda", "alpha"}.
             kl_scale: Initial KL weight scale for lambda LK loss.
             kl_decay: Decay factor for adaptive KL weight in lambda LK loss.
+            gradient_checkpointing: recompute each TTT step in backward instead
+                of retaining its activations (attention probs, MLP intermediates,
+                logits, per-step teacher-p slices). Only the per-step hidden
+                states and cached k/v survive the forward, which is what makes
+                longer max_length fit on a single card. Not supported for usp:
+                it runs collectives inside the step, which is not replay-safe.
         """
         super().__init__()
         self.draft_model = draft_model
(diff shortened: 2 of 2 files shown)


jiapingW:
Hi, the current `main` branch has a huge update. Can you update your PR to match the latest `main`? Thanks a lot.

notwitcheer:
Thanks for the ping. I checked the new `main` before starting the rebase, and none of the code this PR touches is there any more.

Both files moved branches:

- `scripts/train_eagle3.py` is not on `main`, which no longer carries any `train_*.py` entrypoint. It is on `colocated` at `d476d52d`.
- `specforge/core/eagle3.py` is not in `main:specforge/core/`. It is on `colocated` at `6d6607d8`.

So this is a branch move rather than a rebase.

The feature itself does not appear to have landed on `main` yet. A code search across the repo for `gradient_checkpointing`, `checkpoint_activations` and `draft_gradient_checkpointing` returns nothing, and the only memory-management call I found in the new EAGLE-3 path is `torch.cuda.empty_cache()` at `specforge/algorithms/eagle3/model.py:307`. I read that from the code on `main` rather than running it, so treat it as unverified on my side.

Two ways forward, and I am happy with either:

1. Retarget this PR to `colocated`, where it applies as written against the code it was tested on.
2. Open a separate PR against `main` that implements the same thing in the new layout, around `specforge/algorithms/eagle3/model.py` and `specforge/training/trainer.py`.

If the disaggregated path handles activation memory differently and checkpointing is deliberately out of scope on `main`, tell me and I will keep this on `colocated` only. Which would you prefer?

jiapingW:
We think the first way is more suitable.

> Thanks for the ping. I checked the new `main` before starting the rebase, and none of the code this PR touches is there any more.
> 
> Both files moved branches:
> 
> * `scripts/train_eagle3.py` is not on `main`, which no longer carries any `train_*.py` entrypoint. It is on `colocated` at `d476d52d`.
> * `specforge/core/eagle3.py` is not in `main:specforge/core/`. It is on `colocated` at `6d6607d8`.
> 
> So this is a branch move rather than a rebase.
> 
> The feature itself does not appear to have landed on `main` yet. A code search across the repo for `gradient_checkpointing`, `checkpoint_activations` and `draft_gradient_checkpointing` returns nothing, and the only memory-management call I found in the new EAGLE-3 path is `torch.cuda.empty_cache()` at `specforge/algorithms/eagle3/model.py:307`. I read that from the code on `main` rather than running it, so treat it as unverified on my side.
> 
> Two ways forward, and I am happy with either:
> 
> 1. Retarget this PR to `colocated`, where it applies as written against the code it was tested on.
> 2. Open a separate PR against `main` that implements the same thing in the new layout, around `specforge/algorithms/eagle3/model.py` and `specforge/training/trainer.py`.
> 
> If the disaggregated path handles activation memory differently and checkpointing is deliberately out of scope on `main`, tell me and I will keep this on `colocated` only. Which would you prefer?

notwitcheer:
Retargeted to `colocated` per your preference.

I checked this before making the base change. Merging this branch into `colocated` at `40d8fef` produces no conflicts across both files:

- `specforge/core/eagle3.py` is byte-identical on `colocated` to the revision this was developed against (`6d6607d8`), so the TTT unroll change applies as written.
- `scripts/train_eagle3.py` has moved since (`a121fbcc` to `d476d52d`), but the hunks auto-merge cleanly. I checked where they landed: the `--draft-gradient-checkpointing` argument is in `optimization_group`, and the flag still reaches `OnlineEagle3Model` through the existing kwargs. The QwenVL guard sits ahead of model construction as before.

The merged tree compiles clean and the diff is unchanged from what you reviewed.

Ready for review whenever it suits you.

jiapingW:
Can you add the test results of using draft-gradient-checkpointing?

jiapingW (approved):

notwitcheer:
Test results below. The equivalence numbers were in the PR description, so I have also
built the CI test I offered there, which is probably the more useful half of the answer now.

### The test

`tests/test_runtime/test_equiv_draft_grad_ckpt.py`, following the `test_equiv_*` pattern in
`tests/test_runtime/`. It builds a tiny synthetic draft head from the existing
`tests/test_runtime/_fixtures.py`, so there is no model download, and it is
`skipUnless(torch.cuda.is_available())`. Two cases: the bitwise equivalence gate, and that
`usp` is rejected with a `ValueError` rather than silently ignored.

The equivalence bar is bitwise rather than a tolerance. A replay that appends to a shared KV
cache appends a second time and corrupts every later step, and a loose tolerance would hide
exactly that. So the test measures the backward noise floor first, by running the unpatched
path twice on identical inputs, then requires the checkpointed-vs-unpatched gradient
difference to sit at or below that floor.

Run against the merged `colocated` tree at `1879643`, on an RTX 5090 with torch 2.11.0+cu130:

```
[grad-ckpt equiv] noise floor 0.000e+00 | ckpt-vs-base grad diff 0.000e+00 | retained 8.9 -> 0.2 MiB
Ran 2 tests in 4.070s
OK
```

Three consecutive runs gave identical output. Losses, acceptance rates and every parameter
gradient are bitwise equal, and the gradient keys match, so no parameter silently drops out
of the graph. pre-commit is clean.

### Memory

The retained-graph saving scales with sequence length and draft vocab, so the CI fixture
understates it. Measured on the same tree, TTT length 4, retained memory across the forward
with the flag off versus on:

| hidden | vocab | draft vocab | seq | off | on | delta |
|---|---|---|---|---|---|---|
| 64 | 256 | 64 | 64 | 8.9 MiB | 0.1 MiB | -98.5% |
| 512 | 32768 | 8192 | 256 | 44.0 MiB | 18.8 MiB | -57.3% |
| 512 | 32768 | 8192 | 1024 | 177.7 MiB | 74.4 MiB | -58.2% |
| 512 | 32768 | 8192 | 2048 | 351.0 MiB | 148.7 MiB | -57.6% |
| 1024 | 65536 | 16384 | 2048 | 700.9 MiB | 296.7 MiB | -57.7% |

The first row is the CI fixture config. The test's own line reports 0.2 MiB rather than 0.1
for it, which is allocator granularity at that size.

One measurement note in case it saves someone time: this has to be read as a delta across
the forward, not as absolute `torch.cuda.memory_allocated()`. I compared absolutes on my
first pass and the tiny fixture appeared to get worse under checkpointing, because the
weights and the allocator state dominate at that scale.

### Production evidence

From the run in the PR description, not re-measured today: 7,000 steps of online training at
max-length 2048 on one RTX 5090, 36B w4a16 teacher served by the sglang backend, rc=0. The
retained graph dropped roughly 85% there, at TTT 7 with a 155K target vocab and a 32K draft
vocab. Step time went from about 3.5 s at 1024 to about 3.7 s at 2048, so the extra draft
forward sits under a teacher-bound step. Before the patch, 2048 missed single-card fit by
around 250 MB with every other memory option already enabled.

### Where to put it

Since this PR is merged, I will open a follow-up PR against `colocated` with the test file.
Tell me if you would rather it sat elsewhere in the tree, or if you want the memory sweep
included as a second test rather than kept out of CI. I left the sweep out because it takes
GPU minutes and asserts a scale-dependent number, which felt wrong for a gate.

PR #670, opened 2026-07-12
feat(training): `--vocab-mapping-path` to pin the vocab mapping across resumes

notwitcheer:
## The failure mode

The draft-vocab mapping (top `draft_vocab_size` of the target vocab by training-set frequency) is generated with a cache key that includes `max_length` and the other dataset parameters. Resume a checkpoint with any of those changed (e.g. continuing a 1024 run at 2048) and the trainer regenerates a shifted d2t/t2d (the truncation point moves, so tail frequencies reshuffle), then loads it after the head's weights: slot i no longer means the same target token, supervision is scrambled for every shifted slot, and there is no error. Training continues with a plausible loss curve.

## What this does

`--vocab-mapping-path` bypasses generation and uses the given mapping file. Default off; when unset, behavior is unchanged. A missing file raises `FileNotFoundError` up front rather than falling back to regeneration.

## Verification

Resumed a 7,000-step EAGLE-3 head at doubled max-length (1024 -> 2048) with the night-1 mapping pinned, trained another 7,000 steps, and compared a fixed spec-decode eval (same engine config, same prompts, greedy) before and after: no acceptance collapse through the resume (the code workload held at 1.29x speedup / 1.65 accept-len at both points).

## Related

#534 preserves the mapping when fine-tuning from a checkpoint; this flag covers the resume-with-changed-data-parameters case explicitly and gives scripts a way to state the dependency.

### DIFF
--- a/scripts/train_eagle3.py
+++ b/scripts/train_eagle3.py
@@ -473,6 +473,12 @@ def sanity_check(args: Namespace) -> None:
             raise ValueError("shard_target_output is only supported for sglang backend")
         if args.is_vlm:
             raise ValueError("shard_target_output is only supported non vlm model")
+    if args.vocab_mapping_path and not os.path.isfile(args.vocab_mapping_path):
+        # validated here, on every rank, before any distributed barrier:
+        # raising inside rank_0_priority() would hang the other ranks
+        raise FileNotFoundError(
+            f"--vocab-mapping-path {args.vocab_mapping_path} does not exist"
+        )
 
 
 def sp_sanity_check(args: Namespace) -> None:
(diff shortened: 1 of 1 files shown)


notwitcheer:
Both review points addressed in the latest commit: the existence check now runs in `sanity_check` on every rank before any distributed barrier (the raise inside `rank_0_priority()` is gone), and `build_eagle3_dataset` is now skipped when running offline with a pinned mapping — its only offline consumer was `generate_vocab_mapping_file`. The offline skip is logic-verified; our hardware runs exercise the online path.

jiapingW:
Hi, the current `main` branch has a updated and now support disggregation training and inference. The colocated training are now stored in `colocated` branch. Can you help update your PR to currenct `main` and verify its validity. We would be very grateful. Thanks. @notwitcheer

notwitcheer:
Thanks. I checked `main` before rebasing, and there are two findings, the second of which makes this PR unnecessary there.

First, the patch cannot apply. Its only file, `scripts/train_eagle3.py`, is not on `main`; it is on `colocated` at `d476d52d`. `main:scripts/` carries no `train_*.py` entrypoint now.

Second, `main` already has this feature. `vocab_mapping_path` is a config field in `specforge/config/schema.py` and it is honoured in `specforge/training/assembly.py`. It is documented in `docs/basic_usage/training.md` and set across the example configs. The relevant guard returns before the mapping is regenerated:

```python
if (
    cfg.model.vocab_mapping_path
    or bundle.draft_vocab_size == bundle.target_vocab_size
):
    return
```

That is the same semantics this PR adds to the old script, so there is nothing here to port. I read this from the code on `main` rather than running it, so treat the behavioural claim as unverified on my side.

One observation while I was in there. The dataset identity that keys regeneration still includes `max_length` (`assembly.py:442`), and `count_effective_feature_tokens` truncates to `max_length` before counting. A resume with a changed `max_length` and no `vocab_mapping_path` set will therefore still recount and produce a shifted map, which is the failure this PR was written against. That matches the old default-off behaviour and may well be intended now that the guard exists for anyone who wants it. Flagging it in case the default is worth revisiting.

So I would close this one, or retarget it to `colocated` if you want `scripts/train_eagle3.py` to carry the flag there too. Tell me which and I will do it.

jiapingW:
I suggest that you can close thie PR.  Thanks a lot.

notwitcheer:
Closing as suggested.

For anyone finding this later: `vocab_mapping_path` is already a first-class config field on `main` (`config/schema.py`, documented in `docs/basic_usage/training.md`), and `specforge/training/assembly.py:425-429` returns early when it is set, before regeneration. Same semantics this PR was reaching for, so there is nothing to port.

There is one observation worth keeping when this closes. `assembly.py:442` still folds `max_length` into the regeneration identity, so a run that changes only `max_length` regenerates the mapping on the default unpinned path. That is the regeneration this PR originally guarded against, and it survives for users who do not set `vocab_mapping_path`. Happy to open a separate issue if that is useful, or leave it here if it is already known.

Thanks for the quick turnaround on the restructure questions.

Issue #380, opened 2025-12-22
[Bug] Question Regarding Memory Consumption


ggg-s:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/SpecForge/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

The target model is Qwen3-32B, and the log:

[mem][after_init_distributed] allocated=0.0MiB reserved=0.0MiB max_allocated=0.0MiB max_reserved=0.0MiB
[mem][after_build_draft_model] allocated=2884.4MiB reserved=2898.0MiB max_allocated=2884.4MiB max_reserved=2898.0MiB
[mem][after_build_target_model] allocated=11575.9MiB reserved=11610.0MiB max_allocated=11575.9MiB max_reserved=11610.0MiB
[mem][after_build_dataloaders] allocated=11575.9MiB reserved=11610.0MiB max_allocated=11575.9MiB max_reserved=11610.0MiB
[mem][after_load_vocab_mapping] allocated=11575.9MiB reserved=11610.0MiB max_allocated=11575.9MiB max_reserved=11610.0MiB
[mem][after_fsdp_wrap] allocated=9063.3MiB reserved=14842.0MiB max_allocated=14806.8MiB max_reserved=14842.0MiB
[mem][after_forward] allocated=12671.3MiB reserved=20902.0MiB max_allocated=17686.7MiB max_reserved=23330.0MiB
[mem][after_backward] allocated=9430.3MiB reserved=21750.0MiB max_allocated=17686.7MiB max_reserved=23330.0MiB

Why does the GPU memory usage on each card suddenly spike from approximately 15 GB to around 23 GB the moment training starts?

### Reproduction

CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 torchrun --nproc_per_node 8 \
scripts/train_eagle3.py \
    --target-model-path /models/Qwen3-32B \
    --train-data-path dataset/qwen3_distill_dataset.jsonl \
    --output-dir outputs/Qwen3-32B-eagle3 \
[...]

PR #671, opened 2026-07-12
feat(training): single-GPU online-mode memory options

notwitcheer:
## Motivation

Online EAGLE-3 training with a large quantized teacher can fit on one consumer card, but the training side wastes VRAM in four places that matter at 32 GB. With these options, a 36B w4a16 teacher (sglang backend) plus draft training co-reside on one RTX 5090. Related memory issues: #466, #380.

## What this does

Two opt-in flags (default off, no behavior change when off):

- `--optimizer-cpu-offload`: fp32 master params and AdamW state live in pinned host memory and the optimizer step runs on CPU; only bf16 params and grads stay in VRAM. Saved ~8.4 GB for a 1-layer draft head over a 155K vocab. The CPU step cost (~0.5 s) is amortized by the teacher-bound step.
- `--draft-embedding-cpu`: the frozen draft input embedding stays in host memory; it is used for one lookup per batch, and `embed_input_ids` runs the lookup where the table lives. Saved ~1.6 GB.

Two structural changes:

- The draft model is now built on CPU and moved to GPU after the sglang engine has profiled free memory, so `--mem-frac` no longer needs headroom for the draft weights (engine floor 0.784 -> 0.710 in our runs). End state is identical; only the init order changes.
- At `world_size == 1` the FSDP wrap is skipped: FSDP's flat parameter groups allocate gradient storage even for frozen params (a full vocab x hidden grad for the frozen embedding, ~1.5 GB) and provide no sharding value on one rank. `save_checkpoints` handles wrapped and unwrapped models.

One metric fix that becomes necessary at longer sequences:

- The acceptance-rate metric's fp32 softmax over the draft vocab is computed in token chunks, so its peak memory stays flat in sequence length instead of materializing `seq x draft_vocab` fp32 at once.

## Verification

Multiple 7,000-step online training nights on one RTX 5090 (36B w4a16 teacher, max-length 1024 and 2048), rc=0, resume verified across nights; the resulting checkpoints serve normally under sglang EAGLE-3. The numbers above come from `torch.cuda` allocator stats, not nvidia-smi.

## Notes

- The no-FSDP-at-ws1 change affects anyone running world_size 1 today: they get the same model without the wrap. State-dict save/load paths for both cases are exercised by the resume flow.
- pre-commit clean.

### DIFF
--- a/scripts/train_eagle3.py
+++ b/scripts/train_eagle3.py
@@ -1052,6 +1052,12 @@ def main():
     # engine has profiled free memory, so the engine mem fraction does not
     # have to absorb the draft weights (single-GPU online mode).
     draft_model = draft_model.cuda()
+    if args.draft_embedding_cpu or args.optimizer_cpu_offload:
+        if dist.is_initialized() and dist.get_world_size() > 1:
+            raise ValueError(
+                "CPU offloading options (--draft-embedding-cpu, --optimizer-cpu-offload) "
+                "are only supported in single-GPU mode (world_size=1)."
+            )
     if args.draft_embedding_cpu:
         # Frozen vocab x hidden table, used for one lookup per batch; keep it
         # in host memory to free ~2 x hidden x vocab bytes of VRAM
--- a/scripts/train_eagle3.py
+++ b/scripts/train_eagle3.py
@@ -4,6 +4,7 @@
 # for them: VLM (--is-vlm), USP sequence parallelism (--attention-backend usp),
 # the eval loop (--eval-*-path), --resume, and experiment trackers (--report-to).
 import argparse
+import contextlib
 import hashlib
 import math
 import os
@@ -184,6 +185,16 @@ def build_parser() -> ArgumentParser:
         type=int,
         default=7,
         help="The length for Test-Time Training (TTT).",
+    )
+    training_group.add_argument(
+        "--draft-embedding-cpu",
+        action="store_true",
+        help="Keep the frozen draft input embedding in host memory (single-GPU online mode).",
+    )
+    training_group.add_argument(
+        "--optimizer-cpu-offload",
+        action="store_true",
+        help="Keep fp32 master params and AdamW state in host memory (single-GPU online mode).",
     )
     training_group.add_argument("--resume", action="store_true")
     training_group.add_argument(
@@ -527,13 +538,13 @@ def build_draft_model(args: Namespace) -> Tuple[AutoDraftModelConfig, nn.Module]
             draft_model_last_checkpoint,
             attention_backend=args.attention_backend,
             torch_dtype=torch.bfloat16,
-        ).cuda()
+        )
     else:
         draft_model = AutoEagle3DraftModel.from_config(
             draft_model_config,
             attention_backend=args.attention_backend,
             torch_dtype=torch.bfloat16,
-        ).cuda()
+        )
 
     # Load training state (optimizer, scheduler, epoch, step) for true resume
     resume_state = None
@@ -701,7 +712,12 @@ def save_checkpoints(
         os.makedirs(epoch_output_dir, exist_ok=True)
     dist.barrier()
 
-    with FSDP.state_dict_type(eagle3_model, StateDictType.FULL_STATE_DICT):
+    sd_ctx = (
+        FSDP.state_dict_type(eagle3_model, StateDictType.FULL_STATE_DICT)
+        if isinstance(eagle3_model, FSDP)
+        else contextlib.nullcontext()
+    )
+    with sd_ctx:
         model_state_dict = eagle3_model.state_dict()
         state_to_save = {
             "epoch": epoch,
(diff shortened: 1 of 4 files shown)


notwitcheer:
Applied the world_size guard in `sanity_check` (latest commit): `--optimizer-cpu-offload` and `--draft-embedding-cpu` now raise at world_size > 1 — they assume unwrapped parameters, and FSDP's flat groups would misbehave silently. On the two `non_blocking` suggestions: measured on our runs, the CPU optimizer step (~0.5 s) is fully amortized under a teacher-bound ~3.5 s step, so the copies stay synchronous for correctness-by-default; happy to revisit with a benchmark if maintainers want it. The empty-logits guard in `lk_loss` I've left out to keep the diff focused, but can add it here if preferred.

jiapingW:
Currently, as models become larger, the required draft model size is also gradually increasing. For online training, we've already used training-inflation separation to reduce the GPU memory usage of training tasks. Regarding your requirement for training an Eagle 3 model on a single GPU, I believe it certainly exists. However, I wonder if this would provide sufficient computing power to train on a large amount of data to obtain a good draft model. I'd also like to know if you have this issue on your end. @notwitcheer

notwitcheer:
Yes, I ran into that question directly, and I have numbers on it.

I trained a full EAGLE-3 draft head for Hermes-4.3-36B (Seed-OSS arch) on one RTX 5090 32GB using these patches: 54,000 steps over 8 unattended nights, ~55 GPU-hours, with the w4a16 teacher served by sglang on the same card. Every night exited rc=0 and resume across nights held.

On whether one GPU buys enough compute for enough data: the binding constraint was not compute. The head converged at step 21,000, 39% of the epoch. I dry-ran the release bench there and measured 1.30-1.53x on tree-3-4-8. At epoch end, 33,000 steps and roughly 35 GPU-hours later, the same config measured 1.29-1.57x. The nightly fixed-config accept-len froze at night 3 (code 1.67, repetitive 1.46, prose 1.38) and never moved after that. Over the last 60% of the epoch, neither the nightly smoke nor the release bench could resolve any change.

Additional steps on the same data stopped paying long before the single-GPU budget became the limit.

This head lands at 1.29-1.57x against the 1.7-2.2x precedent band for trained EAGLE-3 heads. The convergence result rules out undertraining as the cause. The arithmetic points at serving regime instead. At accept-len 1.80 a free drafter would give ~1.8x; measured is 1.38x, so about a quarter of the theoretical gain pays for the draft pass, because at batch 1 the verify pass has no batching to amortise against. That is a property of batch-1 single-consumer-GPU serving, not of single-GPU training. The next lever I would pull is the data mix or a cheaper draft pass, not more epochs.

For SpecForge, this means single-GPU online training is sufficient to produce a working head for a model that has none, which is the case I care about. It will not close the gap to multi-GPU published numbers, but that gap does not originate in the training budget.

Both flags in this PR are opt-in and default off, so nothing changes for existing users who do not set them.

On the rebase: I saw your notes on #669 and #670 about the disaggregation restructure. I will rebase this one onto the new main and check whether it belongs there or on `colocated`, since the world_size==1 path may sit differently now.

Full numbers and method: https://github.com/notwitcheer/llm-bench-rig/blob/main/reports/hermes-drafter-epoch1.md

notwitcheer:
Following up on the rebase I committed to above.

Same finding as #669: the code this touches is not on `main`. `scripts/train_eagle3.py` has no `main` equivalent, and the other three files sit in the pre-restructure layout. So this is a branch move rather than a rebase, and since you chose retarget on #669 I have done the same here instead of asking the same question twice.

I ran the merge before changing the base. Merging into `colocated` at `40d8fef` gives no conflicts across all four files, and the merged tree compiles:

- `specforge/core/lk_loss.py` is byte-identical to the revision this was developed against.
- `specforge/optimizer.py` and `specforge/modeling/draft/llama3_eagle.py` have moved, but both hunks auto-merge into place: the `cpu_offload` branch in `BF16Optimizer.__init__`, and the host-memory embedding lookup in `embed_input_ids`.

If you would prefer this reimplemented against the new `main` layout, say so and I will open a separate PR. The world_size guard means this stays inert for multi-GPU users in either case.

jiapingW:
> Following up on the rebase I committed to above.
> 
> Same finding as #669: the code this touches is not on `main`. `scripts/train_eagle3.py` has no `main` equivalent, and the other three files sit in the pre-restructure layout. So this is a branch move rather than a rebase, and since you chose retarget on #669 I have done the same here instead of asking the same question twice.
> 
> I ran the merge before changing the base. Merging into `colocated` at `40d8fef` gives no conflicts across all four files, and the merged tree compiles:
> 
> * `specforge/core/lk_loss.py` is byte-identical to the revision this was developed against.
> * `specforge/optimizer.py` and `specforge/modeling/draft/llama3_eagle.py` have moved, but both hunks auto-merge into place: the `cpu_offload` branch in `BF16Optimizer.__init__`, and the host-memory embedding lookup in `embed_input_ids`.
> 
> If you would prefer this reimplemented against the new `main` layout, say so and I will open a separate PR. The world_size guard means this stays inert for multi-GPU users in either case.

Good!

jiapingW (approved):

PR #672, opened 2026-07-12
test(domino): add single-sample overfit and real serving gates

jianuo-huang:
## Dependency and scope

This PR depends on #674 only for the canonical regeneration entrypoints
referenced by its documentation. If `SOURCE_DATA_PATH` already points to
validated regenerated data, the gates here can run independently of #674.

#675 is a complementary transform for exploding multi-turn reasoning data by
assistant turn. It is not required here because this gate deliberately selects
one single-turn `user`/`assistant` row.

This PR does **not** add data-regeneration implementation, modify SGLang server
core, or include FSDP/chunked-loss memory optimizations.

## Summary

This PR adds a reproducible end-to-end correctness gate for Domino disaggregated
training. It starts from one validated regenerated row, overfits it through the
real capture-server/trainer path, exports the resulting draft checkpoint, and
checks that checkpoint through real SGLang DFLASH chat serving.

Two profiles are included:

| Profile | Data contract | Template | Context | Default topology |
|---|---|---|---:|---|
| `qwen3-8b` | non-reasoning | `qwen` | 512 | capture TP1 + trainer DP7; serving TP1 |
| `qwen3.6-27b` | structured reasoning | `qwen3.5` | 2048 | capture TP1 + trainer DP7; serving TP1 |

### 1. Select one strict training sample

`scripts/create_qwen3_8b_overfit_data.py` runs the real tokenizer and SpecForge
preprocessing stack and selects exactly one successful `user`/`assistant` row.
The selected row must:

- satisfy the profile's reasoning policy (`forbidden` for Qwen3-8B, `required`
  for Qwen3.6-27B);
- contain no raw `<think>` markers;
- have at least 128 visible assistant characters and `2 * block_size` loss
  tokens after preprocessing; and
- fit without truncation.

It also writes the exact prompt/target boundary used by the serving gate.
Historical assistant `reasoning_content` is stripped from that request artifact.

### 2. Overfit through the real disaggregated path

`examples/disagg/run_qwen3_8b_domino_disagg_overfit.sh` requires a fresh work
directory and unused ports, then starts:

- Mooncake;
- a real SGLang target capture server;
- a producer that repeatedly serves the selected prompt; and
- a DP consumer that trains the Domino draft (400 optimizer steps by default).

`train_only_last_turn` is passed through both disaggregated preprocessing and
its cache key. The run passes only if the final logged step is exact, loss is at
most `0.0001`, token accuracy is at least `1.0`, and the final checkpoint exists.

### 3. Export the trained draft

The final checkpoint is converted to an SGLang-compatible HF DFLASH artifact.
The draft embedding is populated from the target model so the exported artifact
can be loaded directly by the real serving path.

### 4. Verify real SGLang DFLASH serving

`examples/disagg/run_qwen3_8b_domino_real_serving_gate.sh` launches SGLang with
`--speculative-algorithm DFLASH` and sends a real `/v1/chat/completions` request.
The gate requires all of the following:

- `/server_info.speculative_algorithm == "DFLASH"`;
- no historical `reasoning_content` in the request;
- choice-level `choices[0].meta_info.spec_accept_length >= block_size`; and
- generated target-prefix match of at least one block.

The complete request, response, server metadata, and assertions are written to
`serving_gate.json` for inspection.

## Usage

After placing a validated regenerated dataset at `SOURCE_DATA_PATH`, run the
8B overfit and real-serving gates together with:

```bash
MODEL_PROFILE=qwen3-8b \
RUN_REAL_SERVING_GATE=1 \
bash examples/disagg/run_qwen3_8b_domino_disagg_overfit.sh
```

Set `MODEL_PROFILE=qwen3.6-27b` for the structured-reasoning profile. Paths,
ports, GPU placement, and thresholds can be overridden through the variables
documented in `docs/examples/qwen3-8b-domino-disagg-overfit.md`.

## Validation

- gate-focused tests: `16 passed`
- Black check on changed Python files
- Ruff/F401 check on changed Python files
- shell syntax checks on both launchers
- Qwen3.6 config JSON validation
- `git diff --check`

### Qwen3-8B real-serving evidence

Previously passed against a real SGLang DFLASH server:

- endpoint: `/v1/chat/completions`
- `speculative_algorithm = DFLASH`
- choice-level `spec_accept_length = 16`
- target-prefix match: 16/16 tokens
- historical `reasoning_content` in request: 0

### Qwen3.6-27B profile smoke evidence

The real tokenizer and preprocessing stack selected source row 128 with 925
rendered tokens and 899 loss tokens, without truncation. Structured reasoning
was preserved in the target while historical reasoning was removed from the
serving request.

A Qwen3.6-27B GPU overfit or real-serving end-to-end pass is **not** claimed by
this PR.

### DIFF
--- a/configs/qwen3.6-27b-domino-full-attention.json
+++ b/configs/qwen3.6-27b-domino-full-attention.json
@@ -1,0 +1,52 @@
+{
+  "architectures": [
+    "DFlashDraftModel"
+  ],
+  "attention_bias": false,
+  "attention_dropout": 0.0,
+  "auto_map": {
+    "AutoModel": "dflash.DFlashDraftModel"
+  },
+  "block_size": 16,
+  "bos_token_id": null,
+  "dflash_config": {
+    "mask_token_id": 248070,
+    "target_layer_ids": [1, 16, 31, 46, 61],
+    "projector_type": "domino",
+    "pure_draft_prefix_len": 1,
+    "emb_dim": 256,
+    "gru_hidden_dim": 1024,
+    "shift_label": false
+  },
+  "dtype": "bfloat16",
+  "eos_token_id": 248044,
+  "head_dim": 128,
+  "hidden_act": "silu",
+  "hidden_size": 5120,
+  "initializer_range": 0.02,
+  "intermediate_size": 17408,
+  "layer_types": [
+    "full_attention",
+    "full_attention",
+    "full_attention",
+    "full_attention",
+    "full_attention"
+  ],
+  "max_position_embeddings": 262144,
+  "max_window_layers": 5,
+  "model_type": "qwen3",
+  "num_attention_heads": 32,
+  "num_hidden_layers": 5,
+  "num_key_value_heads": 8,
+  "num_target_layers": 64,
+  "pad_token_id": 248044,
+  "rms_norm_eps": 1e-06,
+  "rope_scaling": null,
+  "rope_theta": 10000000,
+  "sliding_window": null,
+  "tie_word_embeddings": false,
+  "transformers_version": "5.5.3",
+  "use_cache": true,
+  "use_sliding_window": false,
+  "vocab_size": 248320
+}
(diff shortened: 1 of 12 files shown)


jianuo-huang:
Closing this draft in favor of a replacement PR with model-neutral Domino/DFLASH gate naming, profile-parameterized tests, and no merge dependency on the regeneration PR.

PR #678, opened 2026-07-14
refactor: consolidate training on the unified runtime

maocheng23:
## Review status

This monolithic PR now includes the follow-up correctness and scalability repair
at functional commit `f9cc0db1176eb940f2ce575e472398557ffc6fdf`, with
lint-clean head `2d89fd4359f7146a737d87e90a1799958e61bff4`.

- Baseline: `40d8feffdc7ef77a38278802301981897bb6f250`
- Previously reviewed target: `3c89ddc03a38b2835e74c9f95e219d5bacb9bcf1`
- Functional repair: `f9cc0db1176eb940f2ce575e472398557ffc6fdf`
- Current lint-clean head: `2d89fd4359f7146a737d87e90a1799958e61bff4`
- Current tree: `0d49624c6a7a0fbadd2e58ab2227a43a6d85c275`

The draft decomposition [#681](https://github.com/sgl-project/SpecForge/pull/681)
through [#694](https://github.com/sgl-project/SpecForge/pull/694) describes the
earlier reviewed tree. The `f9cc0db` repair and lint follow-up have not yet been
propagated through that stack, so the stack tip and this PR no longer have
identical trees. Review this monolith head for the current end state.

## What changed

The refactor still consolidates training behind one typed run configuration,
one runtime/data-plane contract, and one `Trainer.fit()` lifecycle for EAGLE3,
P-EAGLE, DFlash, Domino, and DSpark.

The follow-up repair additionally:

- removes the author-local paths and ad-hoc hot-loop profiling changes from
  `b909482`, restores the no-tensor invariant with a cheap primitive fast path,
  and retains the required `_load_input_tools` import;
- repairs Ulysses EAGLE3 position IDs, DP-reduced structured metrics, legacy
  vocabulary-mapping export, DeepSeek-V2 templating, and checked-in data paths;
- fixes Domino's NPU bf16 GRU gradient path and makes speculative generation
  actually apply the trained GRU/projection logits head;
- validates finite accumulation plans early and derives deterministic online
  optimizer horizons from complete global dispatch windows;
- batches RefDistributor SQLite and inbox durability work, keeps multi-node
  SQLite/WAL state node-local, and moves scalable provenance resolution to rank
  zero while preserving legacy checkpoint validation;
- closes server-capture/Mooncake response-loss, provisional-ownership, timeout,
  retry, shutdown, post-close drain, and process-group cleanup holes;
- wires the managed SGLang memory/EP settings and makes the live
  server-capture numerics test an always-run GPU CI gate.

[...]

PR #681, opened 2026-07-15
[PR678 stack 00/14] fix(bench): correct HumanEval scoring

maocheng23:
Part **P00/14** of the review stack replacing monolithic #678.

## Review this layer

**Primary decision:** does HumanEval scoring execute the benchmark's canonical `check(candidate)` contract instead of treating any compilable completion as correct?

- Logical base: `main` at `40d8feffdc7ef77a38278802301981897bb6f250`
- Incremental diff: [P00 only](https://github.com/maocheng23/SpecForge/compare/40d8feffdc7ef77a38278802301981897bb6f250...agent/pr678-p00-humaneval)
- Review order: **first / bottom of stack**
- Commit: `cdb81ab`

## What changes

- Execute HumanEval's defined `check(candidate)` function.
- Carry the prompt inside each accuracy label instead of relying on hidden question ordering/state.
- Add focused regression coverage.

No SpecForge runtime capability changes in this layer.

## Validation

- `2 passed`
- Final stacked tip `f9fc810` has the exact Git tree `feb1bc2014b0...` of corrected #678 head `3c89ddc`.

## Known risk

Generated code still executes in-process without a timeout/sandbox; that pre-existing hardening concern is intentionally outside this split.

> Upper PRs temporarily target `main` because their parent refs live in the fork; each body links the authoritative incremental diff.

### DIFF
--- a/benchmarks/benchmarker/humaneval.py
+++ b/benchmarks/benchmarker/humaneval.py
@@ -49,9 +49,11 @@ def check_code_passes_tests(code: str, test_code: str, entry_point: str) -> bool
         namespace = {}
         # Execute the code (function definition)
         exec(code, namespace)
-        # Execute the test code (which contains assertions)
-        # If no exception is raised, the tests pass
+        # HumanEval test payloads define check(candidate); they do not invoke it.
         exec(test_code, namespace)
+        checker = namespace.get("check")
+        if checker is not None:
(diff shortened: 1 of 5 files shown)


jiapingW:
This PR has been included in [PR 682](https://github.com/sgl-project/SpecForge/pull/682).

PR #688, opened 2026-07-15
[PR678 stack 07/14] feat(training): add orchestration and lifecycle services

maocheng23:
Part **P07/14** of the review stack replacing monolithic #678.

## Review this layer

**Primary decision:** are producer/consumer lifecycle, checkpoint/resume, scheduling, profiling, and tracking services correctly isolated?

- Logical parent: [#687](https://github.com/sgl-project/SpecForge/pull/687) at `998fa21`
- Incremental diff: [P07 only](https://github.com/maocheng23/SpecForge/compare/agent/pr678-p06-launch-topology...agent/pr678-p07-orchestration)
- Commit: `5f98191`
- Review order: after #687

## What changes

- Adds prompt construction, scheduling, capture contracts, and disaggregated lifecycle services.
- Adds backend/checkpoint, profiling, tracking, and vocabulary mapping seams.
- Still does not flip the public entry point.

## Validation

- `26 passed`
- One spawned distributed case deferred on macOS because the child requires unavailable Triton.

- GitHub Actions lint: passed after the lint-clean restack.

## Review note

P07 precedes assembly in the actual ancestry; review control-file lifecycle and step-horizon resolution.

> This draft temporarily targets `main` because the logical parent ref lives in the fork. GitHub's Files tab is therefore cumulative; review the incremental link above. Mark this PR ready only after its parent lands/it is retargeted. Final tip `4d9b08a` is tree-identical to corrected #678 head `3c89ddc`.


specforge/data/preprocessing.py
```diff
@@ -764,6 +764,79 @@ def generate_vocab_mapping_file(
...
+        "target": target,
+        "hidden_state": hidden_state,
+        "input_ids": input_ids,
+    }
+
+
+def process_offline_dflash_sample(
+    raw: Dict[str, torch.Tensor], max_len: int
+) -> Dict[str, torch.Tensor]:
+    """Normalize one prepared DFlash-family feature sample.
```
FrankLeeeee, line 795: I do not see the offline processing functions for other spec algos, i assume they are just not included in this PR and left for future PRs?

specforge/training/strategies/base.py
```diff
@@ -86,6 +85,35 @@ def checkpoint_state_filter(self, state_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+        input_ids, target, loss_mask = target_head.preprocess(
+            input_ids, target, loss_mask
+        )
+        target = target_head(target.to(device))
+        return input_ids.to(device), target, loss_mask.to(device)
+    return input_ids.to(device), target.to(device), loss_mask.to(device)
+
+
 class Eagle3TrainStrategy(DraftTrainStrategy):
     """EAGLE3 TTT strategy wrapping the existing ``OnlineEagle3Model``.
```
FrankLeeeee, line 119: Thinking from a semantic coherence point of view, I think we can migrate this to be part of `specforge.algorithm`.

FrankLeeeee: but we can do it in another PR. let's keep this PR as it is for now.

specforge/data/preprocessing.py
jiapingW: > I do not see the offline processing functions for other spec algos, i assume they are just not included in this PR and left for future PRs?

Yes,now it support eagle3 offline training colocate and disaggregated. We will add support for offline training of dflash-style methods in a future PR.

specforge/training/strategies/base.py
jiapingW: > Thinking from a semantic coherence point of view, I think we can migrate this to be part of `specforge.algorithm`.

Sounds good.

PR #703, opened 2026-07-17
fix: rebuild non-persistent rotary buffers after from_pretrained

julyanghar:
## Motivation

Warm-starting draft training from a checkpoint (e.g. `--ckpt-dir`, or `--resume` from a saved run) immediately diverges to `loss=NaN`.

Root cause: the rotary embedding buffers `inv_freq`, `cos_cached`, and `sin_cached` are registered with `persistent=False` (in `specforge/modeling/draft/llama3_eagle.py`), so they are **not** stored in the checkpoint `state_dict`. When the draft model is loaded via `AutoEagle3DraftModel.from_pretrained(...)`, transformers' meta-device / `low_cpu_mem_usage` loading path leaves these buffers **uninitialized** — NaN once the model is moved to GPU — and every RoPE application produces NaN, so training diverges from the first step. The cold-start path (`from_config`) constructs the rotary module normally and is unaffected.

## Modifications

- `scripts/train_eagle3.py` (`build_draft_model`): after `from_pretrained` and before `.cuda()`, re-run `_init_rope()` on every module that owns a rotary embedding, which rebuilds the non-persistent buffers. `_init_rope` already exists, so no model-code change is needed. One added log line, guarded to print once.
- `tests/test_modeling/test_draft/test_llama3.py`: two regression tests —
  - `test_rotary_buffers_absent_from_state_dict` documents the root cause (the three buffers are absent from `state_dict`).
  - `test_rotary_buffers_rebuilt_after_from_pretrained` does a real `save_pretrained` → `from_pretrained` round-trip, applies the `_init_rope` rebuild, and asserts the buffers are finite and identical to a freshly-initialized reference. (Equality is asserted rather than observing NaN, because CPU-only CI does not reliably reproduce the uninitialized-memory NaN the fix prevents on GPU.)

## Related Issues

None known. (Reproducible on any warm start / resume of a draft model via `from_pretrained`.)

## Accuracy Test

Not applicable to inference accuracy — this changes the training script (`build_draft_model`), not model / kernel / architecture code. The behavioral effect is a correctness fix: warm-start training goes from `loss=NaN` (unusable) to normal convergence. The added unit test verifies the rotary buffers are finite and correct after a checkpoint round-trip.

## Benchmark & Profiling

Not applicable. The fix is a one-time buffer rebuild at model-load time (before training starts); it has no steady-state throughput or latency impact.

## Checklist

### DIFF
--- a/tests/test_modeling/test_draft/test_llama3.py
+++ b/tests/test_modeling/test_draft/test_llama3.py
@@ -116,6 +116,62 @@ class TestLlamaForCausalLMEagle3Loading(unittest.TestCase):
             param2 = dict(model2.named_parameters())[name]
             self.assertTrue(torch.equal(param1, param2))
 
+    def test_rotary_buffers_absent_from_state_dict(self):
+        """The rotary buffers are non-persistent, so a checkpoint never stores them.
+
+        inv_freq / cos_cached / sin_cached are registered with persistent=False, so
+        they are absent from state_dict() and thus from any saved checkpoint. This is
+        the property that makes warm-starting safe (see the next test): there is
+        nothing in the checkpoint that could overwrite the live model's correctly
+        initialized rotary buffers. Historically the opposite was dangerous — loading
+        a meta-device model *as* the live model left these buffers uninitialized and
+        training diverged to loss=NaN.
+        """
+        model = LlamaForCausalLMEagle3(self.config)
+        keys = model.state_dict().keys()
+        for buf in ("inv_freq", "cos_cached", "sin_cached"):
+            self.assertFalse(
+                any(k.endswith(buf) for k in keys),
+                f"{buf} unexpectedly present in state_dict (should be persistent=False)",
+            )
+
+    def test_warm_start_preserves_rotary_buffers(self):
+        """Warm-starting from a checkpoint must not disturb the rotary buffers.
+
+        The training warm-start path builds the live draft model via from_config
+        (so __init__ runs _init_rope and the rotary buffers are valid) and then loads
+        only the checkpoint's draft weights with load_state_dict(strict=False). Since
+        the checkpoint state_dict contains no rotary buffers (previous test), the live
+        model's finite buffers must survive untouched. This pins the invariant against
+        a future regression that goes back to using a from_pretrained model as the
+        live model, which would reintroduce the uninitialized-buffer loss=NaN.
+        """
+        # A checkpoint's draft weights: exactly what warm-start loads. Being
+        # persistent=False, it carries no rotary buffers.
+        checkpoint_state = LlamaForCausalLMEagle3(self.config).state_dict()
+
+        # The live model, built the normal way, with valid rotary buffers.
+        live = LlamaForCausalLMEagle3(self.config)
+        rot = live.midlayer.self_attn.rotary_emb
+        before = {
+            name: getattr(rot, name).clone()
+            for name in ("inv_freq", "cos_cached", "sin_cached")
+        }
+
+        result = live.load_state_dict(checkpoint_state, strict=False)
+
+        # The rotary buffers are not tracked by state_dict at all, so they appear in
+        # neither the checkpoint nor the missing/unexpected key sets.
+        for buf in ("inv_freq", "cos_cached", "sin_cached"):
+            self.assertFalse(any(k.endswith(buf) for k in result.missing_keys))
+        rot = live.midlayer.self_attn.rotary_emb
+        for name, ref in before.items():
+            val = getattr(rot, name)
+            self.assertTrue(
+                torch.isfinite(val).all(), f"{name} became non-finite after warm start"
+            )
+            self.assertTrue(torch.equal(val, ref), f"{name} was modified by warm start")
+
     def test_config_validation(self):
         # A dimensionally-valid config (hidden_size divisible by num_attention_heads
         # so it passes transformers' strict config validation) that is still missing


julyanghar:
Hardened the lookup in 66825ad:

- Replaced the bare `hasattr` with `getattr(module, "_init_rope", None)` +
  `callable(...)`, as suggested.

I intentionally did **not** wrap the call in try/except. This fix exists precisely
because uninitialized rotary buffers cause a *silent* `loss=NaN` a few steps into
training. If `_init_rope()` were to raise and we swallowed it with a log line, the
buffers would stay uninitialized and we'd reintroduce that exact silent-NaN
failure — just harder to trace. Fail-fast at startup is the safer behavior here.

I did add loudness where it was actually missing: if *no* rope-owning module is
found, we now emit a warning (that path used to be completely silent). The
regression test mirrors the `callable` check and asserts a module was found.

jiapingW:
Hi, can you help fix the conflicts? We also want you to give a example that can occur ` loss=NaN`. Thanks.

julyanghar:
Thanks for looking at this!

**On the conflict** — it isn't a textual conflict I can rebase away: this PR patches `scripts/train_eagle3.py`, which was removed in d5883d5 ("refactor: consolidate training on unified runtime"). So the fix as written no longer has a home.

**More importantly, I believe that refactor already makes this bug impossible**, so this PR may simply be obsolete. On current `main`:

- `build_eagle3_draft` creates the live model with `AutoDraftModel.from_config(...)`, so `__init__` runs `_init_rope()` and the rotary buffers are correctly initialized.
- `warm_start_draft_model` only does `model.load_state_dict(state, strict=False)`, and `state` comes from a *temporary* `from_pretrained` model whose `state_dict()` — precisely because these buffers are `persistent=False` — contains no rotary buffers at all. That temporary model is then `del`'d.

So the live model's correctly-initialized buffers are never overwritten. The old code was vulnerable only because the meta-loaded model **was** the live model.

**The NaN example you asked for.** This exercises the old path (a `from_pretrained` model used directly as the live model). Output from a single GPU run:

```
=== reference (from_config / __init__) ===
  inv_freq: finite=True    cos_cached[0]=1.000000    sin_cached[0]=0.000000

=== checkpoint state_dict ===
  rotary buffers present: []          # persistent=False -> never saved

=== after from_pretrained(...).cuda() ===
  inv_freq:    finite=True  matches_reference=False  first=2.8698592549372254e-42
  cos_cached:  all zeros    (should be cos(0)=1)
  sin_cached:  all zeros

=== forcing a cos/sin rebuild from the corrupted inv_freq ===
  (any sequence longer than the bogus cached length triggers _set_cos_sin_cache)
  cos: finite=False  nan=4164/266752
  sin: finite=False  nan=4164/266752
  forward output: finite=False  nan=8192
  loss = nan
```

Note that `inv_freq` came back as `-1.13e+27` on one run and `2.87e-42` on the next — it is uninitialized memory, so the symptom is nondeterministic: sometimes the garbage looks benign (RoPE silently degenerates to cos=0 / sin=0, i.e. positional information is destroyed but nothing crashes), and sometimes it produces outright NaN once the cache is rebuilt for a longer sequence. That nondeterminism is what made this hard to track down in the first place.

Repro script:

<details>
<summary><code>repro_rope_nan.py</code></summary>

```python
import tempfile

import torch
from transformers import LlamaConfig

from specforge.modeling.draft.llama3_eagle import LlamaForCausalLMEagle3

cfg = LlamaConfig(
    hidden_size=1024, intermediate_size=2048, num_hidden_layers=1,
    num_attention_heads=8, num_key_value_heads=2, vocab_size=4096,
    draft_vocab_size=1024, max_position_embeddings=2048,
    rms_norm_eps=1e-5, tie_word_embeddings=False,
)
d = tempfile.mkdtemp()

ref = LlamaForCausalLMEagle3(cfg)
ref_rot = ref.midlayer.self_attn.rotary_emb
for n in ("inv_freq", "cos_cached", "sin_cached"):
    t = getattr(ref_rot, n)
    print(f"ref {n}: finite={torch.isfinite(t).all().item()} first={t.flatten()[0].item():.6f}")

sd = ref.state_dict()
print("rotary buffers in state_dict:",
      [k for k in sd if any(k.endswith(b) for b in ("inv_freq", "cos_cached", "sin_cached"))])

ref.save_pretrained(d)
loaded = LlamaForCausalLMEagle3.from_pretrained(d).cuda()
got_rot = loaded.midlayer.self_attn.rotary_emb
for n in ("inv_freq", "cos_cached", "sin_cached"):
    t = getattr(got_rot, n)
    print(f"loaded {n}: finite={torch.isfinite(t).all().item()} "
          f"matches_ref={torch.allclose(t.float().cpu(), getattr(ref_rot, n).float().cpu())} "
          f"first={t.flatten()[0].item()}")

# force a cos/sin cache rebuild from the corrupted inv_freq
long_len = int(got_rot.max_seq_len_cached) + 16
dummy = torch.zeros(1, 1, long_len, cfg.hidden_size // 8, device="cuda")
with torch.no_grad():
    cos, sin = got_rot(dummy, seq_len=long_len)
print(f"after rebuild: cos finite={torch.isfinite(cos).all().item()} nan={torch.isnan(cos).sum().item()}")

x_emb = torch.randn(1, 8, cfg.hidden_size, device="cuda")
x_hid = torch.randn(1, 8, cfg.hidden_size * 3, device="cuda")
attn = torch.ones(1, 8, device="cuda")
with torch.no_grad():
    out = loaded(inputs_embeds=x_emb, hidden_states=x_hid, attention_mask=attn)
    loss = loaded.compute_logits(out).float().log_softmax(-1).mean()
print(f"forward finite={torch.isfinite(out).all().item()}  loss={loss.item()}")
```

</details>

**How would you like to proceed?** As I see it:

1. Close this PR as obsoleted by the refactor.
2. Keep only the regression test, retargeted at the invariant that actually makes the new design safe: that the rotary buffers are absent from `state_dict()`, and that a warm start leaves the live model's buffers intact. That would catch a future change that goes back to using a `from_pretrained` model as the live one.

I'm happy with either — (2) if you think that invariant is worth pinning down, (1) otherwise.

wiketool:
I also encountered this issue during training. When starting training from checkpoint, there would be problems related to rope buffer.

julyanghar:
Following up — I've gone with option (2): the PR is now just the regression tests, with the obsolete fix dropped.

`scripts/train_eagle3.py` no longer exists on `main` (removed in the unified-runtime refactor), so the fix has nowhere left to live — and that same refactor is what makes the bug unreachable now: warm-start builds the live model with `from_config` and only ever loads weights via `load_state_dict(strict=False)`, so the `from_pretrained`-as-the-live-model path that caused it is gone. (I walked through this in the previous comment.)

So the diff is now a single file — `tests/test_modeling/test_draft/test_llama3.py`, `+56` lines — pinning the two properties that safe path relies on:

- **`test_rotary_buffers_absent_from_state_dict`** — the rotary buffers (`inv_freq`, `cos_cached`, `sin_cached`) are registered `persistent=False`, so a checkpoint's `state_dict()` never contains them. Fails if anyone ever flips them to `persistent=True`.
- **`test_warm_start_preserves_rotary_buffers`** — build the model normally (so `_init_rope` gives it valid buffers), then load a buffer-less checkpoint with `load_state_dict(strict=False)`; asserts the live model's rotary buffers stay finite and unchanged. Fails if the load path ever starts clobbering them.

These are unit-level — they construct the model directly rather than driving the full training runtime, so they guard the invariant rather than integration-testing the warm-start path. But they lock down exactly the two things that, if either silently regressed, would bring back the uninitialized-buffer `loss=NaN`.

Ready to merge — or to close if you'd rather not carry the tests. Either's fine with me; thanks for the review!

jiapingW:
Thanks, I'll review it now.

jiapingW (approved):

jiapingW:
LGTM. @julyanghar. Thanks for your observation.

PR #711, opened 2026-07-20
Defer durable DP acknowledgements

zou3519:
## Summary
- add a configurable optimizer-step interval for durable DP acknowledgements
- keep accumulation-window samples separate so unstepped microbatches are never acknowledged
- flush pending acknowledgements before checkpoints, early return, epoch completion, and final completion
- preserve existing behavior with the default interval of 1

## Test plan
- `pytest -q tests/test_runtime/test_trainer.py` (29 passed)
- `pre-commit run --files specforge/launch.py specforge/training/controller.py specforge/training/trainer.py tests/test_runtime/test_trainer.py`


zou3519:
sorry not ready for review yet

PR #714, opened 2026-07-21
DFlash/Domino: Add support for the FlexAttention FLASH backend

zou3519:
Adds an option to turn on FAv4 backend for FlexAttention (it remains off by default). This is significantly faster on Blackwell than the default backend (which is triton). We unfortunately need to monkey-patch PyTorch (in 2.11, which is what SGLang is pinned to right now). Whenever SpecForge can upgrade to PyTorch >= 2.13 then we can get rid of the monkeypatches (and also the correctness test).

This PR also cleans up the attention implementation a little.

### DIFF
--- a/specforge/algorithms/common/dflash_family_model.py
+++ b/specforge/algorithms/common/dflash_family_model.py
@@ -9,6 +9,7 @@ import torch.nn.functional as F
 
 from specforge.core.chunking import checkpointed_chunk_reduce
 from specforge.modeling.draft.dflash import DFlashDraftModel
+from specforge.modeling.draft.flex_attention_backend import flex_attention_backend
 
 try:
     from torch.nn.attention.flex_attention import BlockMask, create_block_mask
@@ -94,6 +95,7 @@ def create_dflash_block_mask(
     S: int,
     block_size: int,
     device: torch.device,
+    flex_block_size=None,
     sliding_window: Optional[int] = None,
 ):
     """Construct a full or sliding Flex Attention mask for DFlash training."""
@@ -128,8 +130,17 @@ def create_dflash_block_mask(
     Q_LEN = N * block_size
     KV_LEN = S + N * block_size
 
+    kwargs = {}
+    if flex_block_size is not None:
+        kwargs["BLOCK_SIZE"] = flex_block_size
     return create_block_mask(
-        dflash_mask_mod, B=B, H=None, Q_LEN=Q_LEN, KV_LEN=KV_LEN, device=device
+        dflash_mask_mod,
+        B=B,
+        H=None,
+        Q_LEN=Q_LEN,
+        KV_LEN=KV_LEN,
+        device=device,
+        **kwargs,
     )
 
 
@@ -309,6 +320,12 @@ class OnlineDFlashModel(nn.Module):
             "block_size": self.block_size,
             "device": device,
         }
+        if (
+            self.attention_backend == "flex_attention"
+            and flex_attention_backend() == "FLASH"
+        ):
+            # FLASH requires a minimum of this block size.
+            mask_args["flex_block_size"] = (256, 128)
         full_attn_mask = mask_builder(**mask_args)
         sliding_window = self.draft_model.sliding_window
         dflash_attn_mask = full_attn_mask
--- a/specforge/algorithms/common/dflash_family_model.py
+++ b/specforge/algorithms/common/dflash_family_model.py
@@ -54,6 +54,9 @@ def create_dflash_sdpa_mask(
     sliding_window: Optional[int] = None,
 ):
     """Construct a full or sliding dense boolean DFlash mask."""
+
+    if sliding_window is not None and sliding_window <= 0:
+        raise ValueError("sliding_window must be > 0")
     B, N = anchor_positions.shape
     Q_LEN = N * block_size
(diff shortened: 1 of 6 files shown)


zou3519:
cc @maocheng23

jiapingW:
@zou3519  Great work! We'll review it soon.

zou3519:
@jiapingW thank you! After my current three PRs (https://github.com/sgl-project/SpecForge/pull/714 https://github.com/sgl-project/SpecForge/pull/715 https://github.com/sgl-project/SpecForge/pull/716) the plan is to 1) get rid of the final blocking syncs (the disagg trainer acks are causing blocking syncs, but I don't know enough about them right now to say how to get rid of them) and then 2) attempt to use torch.compile for the rest of the kernels (I want to do some profiling after those three PRs goes in to see if torch.compile is useful here, it was when I originally tested)

jiapingW:
> @jiapingW thank you! After my current three PRs (#714 #715 #716) the plan is to 1) get rid of the final blocking syncs (the disagg trainer acks are causing blocking syncs, but I don't know enough about them right now to say how to get rid of them) and then 2) attempt to use torch.compile for the rest of the kernels (I want to do some profiling after those three PRs goes in to see if torch.compile is useful here, it was when I originally tested)

Well. Good idea.

gerayking:
I tested this PR at commit `bb4503f` on an NVIDIA B300 SXM6 AC (SM 10.3) with PyTorch `2.11.0+cu130`, CUDA 13.0, and BF16 inputs.

### Performance

The baseline here is FlexAttention with `kernel_options={"BACKEND": "TRITON"}`. The candidate is the new `FLASH` (FA4/CuTeDSL) backend. Both runs use the same logical DFlash mask and include attention forward + backward.

Test shape:

- batch size: 1
- context length: 32,768
- query length: 8,192 (512 draft blocks x 16 tokens)
- KV length: 40,960
- query/KV heads: 32/4
- head dimension: 128
- warmup: 2 iterations; measured: 10 iterations

| Backend | Median | P95 | Compile + warmup | Incremental peak memory |
| --- | ---: | ---: | ---: | ---: |
| TRITON | 18.48 ms | 18.84 ms | 10.14 s | 641 MiB |
| FLASH | 9.16 ms | 9.40 ms | 57.90 s | 641 MiB |

This is a **2.02x steady-state speedup** for attention forward + backward. The first compile is substantially more expensive for FLASH.

BlockMask construction itself was unchanged: 8.89 ms for the TRITON/default block size and 8.84 ms for the FLASH `(256, 128)` block size. Counting mask construction plus attention gives 27.37 ms vs. 18.00 ms, or about **1.52x**. This is a kernel-level benchmark, not end-to-end training throughput; projections, MLP, communication, and other work will reduce the full-step speedup.

### Correctness and numerical precision

The two tests added by this PR both passed on B300, including forward/backward comparison and the DFlash attention module smoke test.

I also tested an irregular case with `context_len=257`, `Q=80`, `KV=337`, five draft blocks, and two invalid/all-masked blocks. FLASH and TRITON both produced finite outputs and Q/K/V gradients. The FLASH-vs-TRITON maximum absolute differences were:

- output: `0.0078125`
- dQ/dK/dV: at most `9.54e-7`

For a broader precision check, I ran three shapes (aligned, irregular with invalid blocks, and a larger GQA training-style shape), each with three random seeds. BF16 FLASH and TRITON results were compared against an explicit FP64 attention reference using the same quantized inputs and a fixed random upstream gradient.

| Mean relative L2 error vs. FP64 | FLASH | TRITON |
| --- | ---: | ---: |
| Output | 0.2261% | 0.2114% |
| dQ | 0.2969% | 0.2959% |
| dK | 0.2932% | 0.2924% |
| dV | 0.2880% | 0.2880% |

Across all nine runs:

- no output or gradient contained NaN/Inf;
- all-masked query rows were exactly zero in FLASH, TRITON, and the FP64 reference;
- FLASH-vs-TRITON output relative L2 error averaged 0.2749%, with maximum absolute difference `0.0078125` and minimum cosine similarity `0.99999577`;
- gradient precision was effectively tied against FP64.

Overall, the measured B300 speedup is real and the numerical differences look consistent with BF16 reduction-order differences. I did not observe a precision blocker for enabling this backend.

drisspg:
we have made a number of codegen and performance improvements in later versions of pytorch for our flex/flash integration, so I would be excited to see if there is some more juice we can squeeze out of it

jiapingW:
I see that you remove the content of `sliding-window`. So, I restored the `sliding-window` and fixed a bug where the `sliding-window` was merely checked for but not actually utilized. Can you help review it? @zou3519

jiapingW:
@zou3519 Hi, we have checked that the [PR731](https://github.com/sgl-project/SpecForge/pulls) has supported complete `sliding-window` draft model layer type and fix the bug that `anchors -1`.  You can review his PR together. I will merge the FLASH optimizations after it handles over.

maocheng23:
Great work! Let's combine this with PR 731 before final merge.

zou3519:
Sorry I missed the comments, will reply faster next time. #731's bugfixes look good to me. I'll rebase this PR today

zou3519:
@jiapingW changes LGTM, thanks for catching the sliding window thing. I rebased the PR

jiapingW (approved):

PR #715, opened 2026-07-21
Domino/Dflash: avoid blocking syncs

zou3519:
Converting CUDA Tensors to scalars is generally bad for training so we avoid it. If there are no Tensor-to-scalar conversions on the majority of steps (forward+backward+optimizer), we are able to hide some CPU dataloading latency.

There were three main sources of Tensor to scalar conversion or scalar host-to-device syncs that we removed:
1. We used to acquire scalar metrics after each step, even if our log interval was greater than one step. This PR changes it so that we acquire Tensor metrics (that are detached), and only convert them to scalar when needed for the log interval.
2. We used to create an `anchor_positions` tensor with shape `valid_counts.max().item()`. This tensor was padded and masked (with `block_keep_mask`). This PR changes it so the shape is padded to `self.num_anchors`. This may waste more performance but does help us get rid of the `.item()`.
3. We used to create CUDA scalar tensors from Python scalars in the Domino hot path. These tiny host-to-device copies can still introduce stream syncs, so sentinel values now use scalar literals and the `lambda_base` metric is filled on device.

We tested this on GB300 with 3 trainers and 1 rollout. At that setup, the net throughput was roughly equal to before. We will actually see perf gains once all of the blocking syncs are removed (I think the last one after this PR is figuring out the ack situation -- the acks happen per-step and are blocking).


zou3519:
cc @maocheng23

specforge/algorithms/common/dflash_family_model.py
```diff
@@ -474,31 +474,34 @@ def _sample_anchor_positions(
         bs = self.block_size
         bsz = loss_mask.shape[0]
         max_anchor = max(seq_len - bs, 0)
+        max_n = max(1, self.num_anchors)
```
zou3519, line 477: this probably needs careful reading

specforge/algorithms/common/dflash_family_model.py
```diff
@@ -474,31 +474,34 @@ def _sample_anchor_positions(
...
 
         valid = loss_mask[:, : max_anchor + 1] > 0.5
         valid_counts = valid.sum(dim=1)
-        max_n = max(1, min(self.num_anchors, int(valid_counts.max().item()) - 1))
```
zou3519: Btw, is the `int(valid_counts.max().item()) - 1)` intentional? (the -1) thing. It looks like when there are fewer than num_anchor anchors, we just drop one of the anchors.

I tried to replicate this behavior in the diff.

zou3519: nvidia-nemo says the -1 is wrong: https://github.com/NVIDIA-NeMo/Automodel/blob/c37f04cde41dd91e2018622cc31e07d0c1462710/nemo_automodel/components/speculative/dflash/core.py#L249-L255

jiapingW: We have checked the code that DFlash and Domino impl have this question. The -1 is not right when the available training tokens is less. We will fix it. Good findings!

gerayking:
Independent B300 validation on head `3b83d8bd20122d027793eb8b9f0bfb9bd21e6cfe` (PyTorch 2.11.0+cu130, CUDA 13.0):

- Targeted tests passed: 5/5 Domino model tests and 30/30 trainer tests.
- `_sample_anchor_positions` synchronized wall time improved from 224.72 to 155.51 us (1.45x) for dense 4096-token masks and from 202.75 to 154.54 us (1.31x) for sparse masks. These are sampling-only numbers, not end-to-end training throughput.

I found three issues that I think should be addressed before merge:

1. The acknowledged off-by-one remains. With valid counts `[5, 2]`, the PR retains `[4, 2]` anchors because `batch_anchor_limit = valid_counts.max() - 1`. The new `test_anchor_sampling_uses_fixed_width_without_scalar_sync` also asserts `[4, 2]`, so it currently codifies the behavior that the discussion says is wrong.
2. Fixed-width output can substantially expand downstream tensors for sparse/short sequences. In my sparse case, the anchor shape changed from `[3, 127]` at the base to `[3, 512]` at the PR head (4.03x wider); for valid counts `[5, 2]`, it changed from `[2, 4]` to `[2, 512]` (128x). `block_keep_mask` removes invalid blocks from the objective/attention visibility, but position, embedding, mask, and draft-model tensors are still constructed at the padded width. The local sampling speedup therefore does not establish an end-to-end win for these distributions.
3. The shared `OnlineDFlashModel._sample_anchor_positions` path still calls `valid_counts.max().item()` and still creates CUDA scalar tensors from Python values. The sync-free anchor change is only in the Domino override. The metric-conversion change benefits both, but the DFlash anchor hot path remains synchronized.

Recommendation: fix the off-by-one and its test, bound the padded width without a host sync (or demonstrate the full downstream tradeoff with an end-to-end benchmark), and either update the DFlash path or narrow the stated scope.

jiapingW:
Help fix the conflicts. Thanks a lot. @zou3519

jiapingW:
This PR's optimization has been implemented in  [PR](https://github.com/sgl-project/SpecForge/pull/759) and has been merged. Welcome to check it. If you think it's OK, you can close this PR. @zou3519

PR #742, opened 2026-08-01
fix(export): preserve both RoPE scaling schemas

Boreas618:
## What changed

- Mirrors non-default `rope_parameters` and legacy `rope_scaling` schemas after HF or SGLang export when serialization keeps only one form.
- Preserves top-level `rope_theta` when constructing the modern representation.
- Leaves default RoPE and already-compatible configs untouched.
- Adds CPU tests for modern-to-legacy, legacy-to-modern, and no-op behavior.

## Why

Transformers 5 serializes the modern schema, while older Transformers and serving stacks may read only `rope_scaling`. Losing YaRN metadata makes a long-context draft silently fall back to unscaled RoPE and can collapse acceptance beyond the original context.

The original fix lived in the removed legacy DSpark trainer. This PR moves the safeguard to the current export boundary, where runtime checkpoints become loadable model directories.

## Impact

Exported long-context drafts remain readable across current and legacy consumers without changing default-RoPE checkpoints.

## Validation

- `python -m unittest tests.test_runtime.test_export.TestRoPEConfigCompatibility`
- `pre-commit run --files specforge/export/checkpoint_io.py specforge/export/to_hf.py specforge/export/to_sglang.py tests/test_runtime/test_export.py`
- `git diff --check upstream/main...HEAD`

The existing GPU exporter round-trip suite was also launched in this CUDA-enabled checkout; the focused compatibility tests above are the deterministic coverage added by this PR.

### DIFF
--- a/specforge/export/checkpoint_io.py
+++ b/specforge/export/checkpoint_io.py
@@ -59,6 +59,8 @@ def apply_legacy_rope_scaling(output_dir: str) -> bool:
         config["rope_scaling"] = {
             key: value for key, value in rope_parameters.items() if key != "rope_theta"
         }
+        if "rope_theta" in rope_parameters:
+            config["rope_theta"] = rope_parameters["rope_theta"]
     elif (
         rope_scaling
         and not rope_parameters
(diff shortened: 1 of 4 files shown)


maocheng23:
For default, let's also preserve default instead of skipping?

maocheng23:
This would also impact custom theta right?

PR #759, opened 2026-08-08
[Feature] Consumer performance optimize

jiapingW:
## Motivation

Online and disaggregated training for the DFlash model family still contained several avoidable synchronization points on the training hot path:

- The maximum valid anchor width was derived from a CUDA tensor with `.item()`, forcing a device-to-host synchronization before every forward pass.
- Metrics were eagerly converted with `.item()`/`.tolist()` on every micro-step, even when they were not logged yet.
- Training batches used blocking host-to-device copies from pageable memory.
- Failed Mooncake sample removals could remain pending after their read leases had been durably acknowledged.
- PyTorch's automatic FlexAttention backend selection could choose the restrictive Flex Decoding path for short DFlash-family queries, which is not suitable for the sparse block masks used by these objectives.

This PR aims to improve steady-state training throughput while preserving the DFlash, Domino, and DSpark objectives, metric semantics, and durable-ACK guarantees. It also adds runnable Qwen3.6-27B Domino/DSpark recipes for online disaggregated training.

## Modifications

- Optimize the training hot path:
  - Compute `max_valid_anchors` from the CPU loss mask before the batch is transferred to the GPU, and pass the result explicitly to DFlash, Domino, and DSpark model code.
  - Enable pinned-memory batches when CUDA and data-loader workers are available, and use non-blocking host-to-device transfers for training tensors.
  - Keep scalar and structured metrics as tensors until a logging or result-materialization boundary instead of synchronizing every micro-step.
  - Reduce distributed metrics only at optimizer-step boundaries and pack scalar reductions where possible.
  - Keep Domino `lambda_base` telemetry as a host scalar instead of constructing an additional CUDA tensor.

- Make FlexAttention backend selection explicit:
  - Set `kernel_options={"BACKEND": "TRITON"}` for DFlash-family FlexAttention calls at every sequence length.
  - This prevents the automatic backend selector from routing short sparse queries to Flex Decoding and applies consistently to DFlash, Domino, and DSpark.

- Improve disaggregated sample cleanup:
  - Add batched, targeted retries for pending Mooncake sample removals.
  - Trigger non-sleeping retries at optimizer boundaries after durable acknowledgements.
  - Extend cleanup statistics and tests while keeping the training path non-blocking.

- Add and update recipes:
  - Add a Qwen3.6-27B DSpark online-disaggregated configuration.
  - Update the Qwen3.6-27B Domino online configuration.
  - Add a reusable Domino launch script for externally managed services.

- Extend regression coverage for configuration topology, DFlash-family chunking and gradients, distributed trainer parity, Mooncake lifecycle behavior, durable acknowledgements, and process-tree cleanup.

## Related Issues

N/A.

## Accuracy Test

Full test-suite result on four NVIDIA H200 GPUs:

```text
Ran 998 tests in 418.932s
OK (skipped=3, expected failures=1)
```

The three skipped tests require optional real-server or real-Mooncake environments. The expected failure documents an existing Mooncake cross-process abort/lease limitation and is unrelated to this PR.

No formal downstream model-quality or acceptance-length evaluation was run. The performance runs use asynchronous capture and may consume samples in a different order, so their step-by-step losses are not treated as a direct accuracy comparison.

## Benchmark & Profiling

Setup:

- Hardware: 2 x NVIDIA H200 GPUs.
- GPU 1: SGLang capture server.
- GPU 2: Domino trainer.
- Target model: `Qwen/Qwen3.6-27B`.
- Training batch size: 1.
- Gradient accumulation steps: 8.
- Run length: 30 optimizer steps per revision.
- Measurement window: steps 11-30, excluding the first 10 steps for loading, compilation, and initial capture wait.
- Samples in the measured window: 160 per revision.
- Baseline: `origin/main@a6b6b02`.
- PR benchmark revision: `0ac2198` (the subsequent commits only update configurations and tests).

Steady-state results:

| Metric | `origin/main` | This PR | Change |
| --- | ---: | ---: | ---: |
| Optimizer-step time | 1.7391 s | 1.5835 s | -8.95% |
| Training compute time | 1.7290 s | 1.5723 s | -9.06% |
| Data-wait time | 0.0049 s | 0.0010 s | -79.6% |
| Training throughput | 4.6001 samples/s | 5.0521 samples/s | +9.82% |
| Estimated optimizer steps/hour | 2,070 | 2,273 | +9.82% |

The two consecutive ten-step measurement windows showed approximately +10.6% and +9.1% throughput improvements. The average consumed sequence length was 2,009.4 tokens/sample on `origin/main` and 2,020.4 tokens/sample on this PR, so the improvement was not caused by shorter records. The capture phase was also similar (49.34 s versus 48.16 s); component timing attributes most of the improvement to the trainer compute path.

The reported optimizer-step timing ends after metric materialization, so the result includes the deferred synchronization cost rather than merely moving it outside the measured region. No kernel-level Nsight or PyTorch profiler trace, and no isolated FlexAttention microbenchmark, is included in this PR.

## Checklist

- [✅] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [✅] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).


examples/configs/qwen3.6-27b-domino-online.yaml
```diff
@@ -31,11 +32,24 @@ training:
...
+  in_flight_high_watermark: 1024
+  in_flight_low_watermark: 512
+
+
 deployment:
   mode: disaggregated
   trainer:
     nnodes: 1
-    nproc_per_node: 8
+    nproc_per_node: 1
```
maocheng23, line 52: only 1 dp?

examples/disagg/run_qwen3.6-27b-domino.sh
```diff
@@ -0,0 +1,61 @@
+# producer
```
maocheng23, line 1: This file seems not reproducible?

maocheng23 (commented):
Over all LGTM, NITs

examples/disagg/run_qwen3.6-27b-domino.sh
jiapingW: > This file seems not reproducible?

This file is used for reproduce the training effiency of Qwen3.6-27B domino online training on H200 and it contains the full commands that seperate the consumer and producer.

examples/configs/qwen3.6-27b-domino-online.yaml
jiapingW: This is used for training test. I think user can modify it according to their situations.

PR #777, opened 2026-08-20
fix(data): skip zero-message conversations before apply_chat_template

zhentaocc:
## Problem

`GeneralParser.parse` drops leading non-user turns while assembling the message
list. For assistant-first rows with no system prompt, this can leave `messages`
empty. On transformers 5.x, `tokenizer.apply_chat_template` indexes
`conversation[0]` unconditionally and raises `IndexError`.

That `IndexError` is **not** caught by the existing `except (ValueError, TypeError)`
fallback around `apply_chat_template`, so a single bad row aborts the entire
`dataset.map` preprocessing job (observed during `build_eagle3_dataset` /
hidden-states prep on EAGLE3 data containing assistant-first conversations).

## Fix

Before calling `apply_chat_template`, detect the empty-`messages` case and return
an empty `(input_ids, loss_mask)` sample. The downstream trainable-token /
loss-mask eligibility filters (`minimum_valid_tokens`, `loss_mask_filter` in
`preprocessing.py`) already drop zero-length samples, matching the existing
None-source skip path. A warning is emitted so these rows are visible.

## Validation

Reproduced the original `IndexError` traceback during dataset preprocessing on
assistant-first EAGLE3 data; with this guard the offending rows are skipped and
preprocessing completes.


jiapingW:
I think that this PR will be closed because all conversations will have a system prompt. This exception will never happen.@zhentaocc

PR #787, opened 2026-08-28
chore: bump SGLang to 0.5.18

maocheng23:
## Summary

- bump SGLang to 0.5.18 and align the pinned Torch/Transformers versions
- update the CUDA/ROCm image references and installation documentation
- port the online spec-capture and Ascend Mooncake patches to the v0.5.18 source layout
- adapt offline capture to the v0.5.18 ParallelState, MLP-sync, request-range, ForwardBatch, hidden-state, and DP-attention APIs
- retire the superseded v0.5.14 patch target now that SGLang #31847's Inkling support is included in v0.5.18; retain the separate Kimi K3 fork target
- make patch application exact and recover stale patch files left when a cached virtualenv upgrades SGLang
- align the CUDA CI FlashInfer cubin with SGLang 0.5.18

## Comparison with #605

The v0.5.14 base and Ascend patches are now replaced directly by their v0.5.18 ports. GitHub recognizes both as renames:

- base patch: +178/-105 (83% similarity)
- Ascend overlay: +3/-13 (84% similarity)

The whole PR is +693/-361. Excluding those versioned patch deltas, the hand-written integration change is +512/-243 (755 changed lines), slightly less than #605's +627/-139 (766 changed lines).

PR #605 merged before the checked-in online capture patch and live capture gate were added. This upgrade therefore also has to port a 12-file SGLang server patch that #605 did not contain.

## CI fixes

The first PR Test restored a cached `sf` virtualenv containing the v0.5.14 patch record and patch-added sink file. Pip replaced SGLang's package-owned files with v0.5.18 but left those unowned files. The installer then used BSD `patch --reverse`, which ignored the requested direction, accepted fuzz/offsets, partially applied old hunks, and failed before the live test started.

The installer now uses exact, atomic `git apply` checks. It recovers only the recognized pip-upgrade residue when the installed version matches the target and the new patch applies exactly; ambiguous states are restored unchanged and rejected. Regression tests cover recovery, idempotence, failure restoration, and rejection of the removed v0.5.14 target.

The restored venv also retained `flashinfer-cubin==0.6.12` while SGLang 0.5.18 installed `flashinfer-python==0.6.17`. CUDA CI now mirrors upstream SGLang and explicitly installs the matching 0.6.17 cubin. The full GPU suite then exposed two v0.5.18 offline API requirements, now handled by initializing each request's extend range and passing `return_hidden_states_before_norm=False` to `ForwardBatch.init_new`.

## Validation

- resolved the project dependencies for Python 3.11 / Linux / CUDA 13.0 with `sglang==0.5.18`, `torch==2.13.0+cu130`, and `transformers==5.12.1`
- ran every repository pre-commit hook
- applied the base patch to a clean upstream SGLang `v0.5.18` checkout
- applied the Ascend overlay on top of the base patch
- compiled all 12 patched SGLang Python files
- reproduced the cached-v0.5.14-to-v0.5.18 transition against a real SGLang tree
- passed all three patch-installer regression tests, including safe restoration on an unknown state and rejection of the removed v0.5.14 target
- GitHub Lint passed on the latest cleanup commit
- GitHub PR Test passed on the latest cleanup commit, including the live GPU server-capture gate and the full test suite


jiapingW:
LGTM. But when we set `return_logprob=True`, the logprob branch does not propagate `last_hidden_states ` into `LogitsProcessorOutput` . Consequently, capture requests that require last_hidden fail in the sink. This doesn't affect our use, but we might use this parameter during testing.

jiapingW (approved):

PR #783, opened 2026-08-25
feat: colocated online training with in-process SGLang capture

maocheng23:
## Motivation

Stack 4/5 replacing draft #766 — the colocated online runtime itself. SpecForge online training currently requires a disaggregated producer/consumer topology even when target capture and draft training fit on the same accelerator, adding feature serialization, transport, and a second GPU pool. This PR adds a bounded colocated path: every trainer rank owns its local SGLang target shard and FSDP draft shard, captures one target batch on demand, trains its rank-local slice, and releases it before the next capture.

Design goal for this cut: colocated is a **structural peer of disaggregated**, and the disaggregated path is not changed. Every shared-path touch is listed under "Shared paths" below.

## Structure

| Concern | Disaggregated | Colocated (this PR) |
| --- | --- | --- |
| Run assembly | `training/disaggregated.py::build_disaggregated_run` | `training/colocated.py::build_colocated_run` |
| Topology dispatch | `assembly.build_training_run` | same function, topology-only dispatch (`_with_run_logger`, `_build_offline_colocated_run`) |
| Online prompt plan | producer epoch permutation | **shared** `training/prompt_plan.py`: same per-epoch permutation and task ids; islands take strided, equal, batch-aligned shards → colocated vs disaggregated runs over the same prompts/seed are comparable at matched samples |
| `model.sglang_*` → engine args | `launch_plan._sglang_argv` | **shared** `launch_plan.resolve_sglang_engine_args` (CLI renderer and in-process kwargs are two views of one resolution) |
| Feature source | `SGLangServerCaptureAdapter` (Mooncake) | `LocalSGLangCaptureAdapter` over `OfflineSGLangCaptureBackend.capture_rows` (#780) |
| Ref queue | `StreamingRefQueue` | `LocalRolloutStream`: pull-through, lazy prompt stream, stages at most one local batch |
| Trainer | `launch._assemble_trainer` | same |
| Topology builder | `build_disagg_online_*` | `build_colocated_online_runtime` (now in `launch.__all__`; five builders as ARCHITECTURE.md states) |

## Modifications

- `deployment.mode: local_colocated` validates for online runs; colocated checks cover context-length headroom, request/token capacity versus `tp_size * batch_size`, SGLang DP/EP constraints, and the sp=1 limitation. Disaggregated/offline validation is unchanged.
[...]

PR #797, opened 2026-09-01
[fix] small improvements for dataloader, metrics, optimizer robustness, and ops

sherlockwu:
**Robustness**
- `optimizer.py`: a non-finite global grad norm now skips the optimizer step deterministically on every rank (clipping cannot rescale a NaN norm; one such Adam step NaNs the weights permanently).
- `mooncake_store.py`: transient `get_into` failures (`TRANSFER_FAIL`/-800 under source-side contention) are retried with backoff instead of killing the run as a `KeyError`; the receive buffer of a failed attempt is quarantined — a timed-out transfer can still write into its destination after the call returns, and reusing that storage has corrupted the heap. Observed rate: ~9 per 100 steps on long 32K runs; an 8K validation run on this branch without the retry died within 12 steps.

**Performance**
- `feature_dataloader.py`: make `dataloader_num_workers` actually multi-threaded. Previously the value only sized the buffer of finished batches — a single background thread fetched one batch at a time, capping delivery at 1/fetch-latency no matter how deep the buffer. Now the prefetch worker submits fetches to a `num_workers`-sized thread pool and the buffer carries futures, so up to `num_workers` batches materialize concurrently while leasing continues.

Useful for long context feature passing.  On the 32K DSV4-Flash disaggregated recipe (2-rank consumer, 3× TP2 capture, identical stack; steady-state tails):

  | loader | step time | exposed fetch wait / step |
  |---|---|---|
  | one concurrent fetch (the previous design's behavior) | 122.6 s | 92.8 s |
  | this PR, `num_workers: 16` | **31.4 s** | 14.6 s |

- `dflash_family_model.py`: `masked_fill` instead of a per-microbatch pageable H2D copy; a device-side assert instead of a host sync right after a collective.

**Observability**
- `controller.py` + loader perf counters: `perf/data_wait_producer_s`, `perf/data_wait_fetch_s`, `perf/fetch_seconds_per_sample`, `perf/fetch_delivered_gib_per_s` — attributes trainer data waits to "capture cannot keep up" vs "transfer is slow".


specforge/runtime/data_plane/feature_dataloader.py
```diff
@@ -397,20 +453,17 @@ def _worker() -> None:
...
-                        self.queue.fail(
-                            refs, reason=f"materialize:{exc}", retryable=False
-                        )
-                        state.mark_yielded_or_failed(refs)
-                        put_interruptibly(exc)
-                        return
-                    state.mark_materialized(refs)
-                    if not put_interruptibly((batch, refs)):
+                    future = pool.submit(self._timed_make_batch, refs)
+                    state.track_fetch(future)
```
maocheng23, line 457: track_fetch() removes the future as soon as it completes, while mark_materialized() happens only when the consumer calls future.result().

So it would trigger a race?

specforge/runtime/data_plane/mooncake_store.py
```diff
@@ -366,6 +369,23 @@ def _store_put_tensor(self, key: str, t: torch.Tensor) -> None:
...
+    def _fetch_tensor(self, key: str, spec) -> torch.Tensor:
+        """Allocate and fetch, retrying transient transfer failures
+        (TRANSFER_FAIL/-800) into a fresh buffer each attempt."""
+        for attempt in range(3):
+            dst = _alloc_from_spec(spec)
+            try:
+                self._store_get_tensor(key, dst)
+                return dst
+            except KeyError as exc:
+                self._quarantined_buffers.append(dst)
```
maocheng23, line 381: This _quarantined_buffers will not be released?

specforge/runtime/data_plane/mooncake_store.py
```diff
@@ -366,6 +369,23 @@ def _store_put_tensor(self, key: str, t: torch.Tensor) -> None:
...
+                delay = 2.0 * 2**attempt
+                logger.warning("%s; retry %d/3 in %.0fs", exc, attempt + 1, delay)
+                time.sleep(delay)
+        dst = _alloc_from_spec(spec)
+        self._store_get_tensor(key, dst)
```
maocheng23, line 386: Why there is no more try catch?

specforge/runtime/data_plane/mooncake_store.py
```diff
@@ -366,6 +369,23 @@ def _store_put_tensor(self, key: str, t: torch.Tensor) -> None:
...
 
+    def _fetch_tensor(self, key: str, spec) -> torch.Tensor:
+        """Allocate and fetch, retrying transient transfer failures
+        (TRANSFER_FAIL/-800) into a fresh buffer each attempt."""
+        for attempt in range(3):
+            dst = _alloc_from_spec(spec)
+            try:
+                self._store_get_tensor(key, dst)
+                return dst
+            except KeyError as exc:
```
maocheng23, line 380: Should this retry only the explicit transient TRANSFER_FAIL/-800 status?

specforge/optimizer.py
```diff
@@ -139,6 +139,22 @@ def _clip_grad_norm(self):
...
+                    "skipping optimizer step: non-finite global grad norm "
+                    f"(max_grad_norm={self.max_grad_norm})"
+                )
+            with torch.no_grad():
+                for p in self.model_params:
+                    p.grad = None
+                for mp in self.fp32_params:
+                    mp.grad = None
+            self.last_grad_norm = grad_norm.detach()
+            return self.last_grad_norm
```
maocheng23, line 157: (nit)Here we still silently return, so there will be a empty step.

specforge/training/controller.py
```diff
@@ -801,6 +801,31 @@ def _fit(self, data: Iterable[TrainBatch], progress: Optional[Any]) -> int:
...
+                        batches = max(1.0, loader["fetch_batches"])
+                        log_metrics.update(
+                            {
+                                "perf/data_wait_producer_s": (
+                                    loader["wait_producer_s"] / steps
+                                ),
+                                "perf/data_wait_fetch_s": (
+                                    loader["wait_fetch_s"] / steps
+                                ),
+                                "perf/fetch_seconds_per_sample": (
```
maocheng23, line 819: This is per batch not per sample right?
As the denominator is batches

maocheng23:
Measured this branch (merged onto main `c49bab49` + #783/#831/#832) on B300 with Qwen3.8-27B NVFP4 target + DFlash2 draft, `max_length 8192` (~277 MB of hidden states per sample), Mooncake TCP, `dataloader_num_workers: 8`, steady-state windows of 130 optimizer steps:

| layout | main loader (one fetch in flight) | this PR (8 concurrent fetches) |
|---|---|---|
| 8 capture servers on node B + DP8 trainer on node A (cross-node, 800 GbE) | 10.97 samples/s, trainer bubble 35-40%, fetch p50 0.46 s/sample | **13.97 samples/s (+27%)**, bubble 5.6%, exposed fetch wait 25% of rank time, fetch p50 1.10 s/sample |
| 4 servers + DP4 trainer, single node | 7.81 samples/s, bubble 37-42%, fetch p50 0.23 s/sample | 7.63 samples/s, bubble 6%, exposed fetch wait 28%, fetch p50 1.23 s/sample |

So the pool converts the wait into overlapped fetches, but on one node the same ~2 GB/s reaches the trainers either way; the gain shows up when the consumer pulls from another node.

Two operational notes from the same runs: with 8 fetches in flight the default 500 ms and a 3 s Mooncake lease both produced `LEASE_EXPIRED` (-707); the new retry kept the run alive, but the backoff sleeps distort timing, so I ran with `--default_kv_lease_ttl=10000` (harmless once #832 makes removal not lease-bound). And the `perf/data_wait_fetch_s` vs `perf/data_wait_producer_s` split matched the per-segment records from an external tracer, which is a nice confirmation of the attribution.

specforge/runtime/data_plane/feature_dataloader.py
jiapingW: Fixed in 214db78. The future and refs now remain tracked until the consumer atomically claims them or shutdown settles them. Shutdown waits for all tracked futures and treats successfully materialized but unyielded consume-once refs as terminal instead of requeueing them. Added regression coverage for this race.

specforge/runtime/data_plane/mooncake_store.py
jiapingW: Addressed in 214db78. A failed get_into destination cannot be released immediately because a late transfer may still write into it. The quarantine is now bounded (8 GiB by default), exposed through health metrics, and fails loudly at the cap instead of growing without bound.

specforge/runtime/data_plane/mooncake_store.py
jiapingW: Fixed in 214db78. All attempts, including the final one, now use the same exception path; the destination buffer from the final failed attempt is quarantined before the error is propagated.

specforge/runtime/data_plane/mooncake_store.py
jiapingW: Fixed in 214db78. Retries now use a typed Mooncake status and an explicit transient allowlist: TRANSFER_FAIL (-800) and LEASE_EXPIRED (-707). Permanent object/lifecycle errors, missing status values, and short reads are not retried. Tests cover transient and permanent statuses.

specforge/optimizer.py
jiapingW: Fixed in 214db78. A non-finite global grad norm now clears gradients and raises FloatingPointError before Adam or the scheduler advances. The controller therefore cannot increment global_step or durably acknowledge that optimizer window. Added optimizer- and controller-level regression tests.

specforge/training/controller.py
jiapingW: Fixed in 214db78. The loader now records fetch_samples, and fetch_seconds_per_sample divides cumulative fetch time by the number of refs rather than the number of fetched batches. Added a metric regression test.

specforge/runtime/data_plane/mooncake_store.py
```diff
@@ -253,6 +265,7 @@ def __init__(
         max_hold_age_s: Optional[float] = None,
         retain_on_release: bool = False,
         max_release_attempts: int = 3,
+        max_quarantined_bytes: int = _DEFAULT_MAX_QUARANTINED_BYTES,
```
maocheng23, line 268: [NIT]Should we add a config here?

jiapingW: Addressed in d114451. Added typed `runtime.feature_store_max_quarantined_bytes` with the existing 8 GiB default and non-negative validation, then wired it through the unified Mooncake store factory. It is now configurable from YAML or a dotted CLI override without hidden environment state. Added schema/default/validation coverage, a factory-forwarding regression test, and config documentation.

jiapingW (approved):

PR #798, opened 2026-09-01
[feature] deepseek v4 flash dspark GQA drafter training

sherlockwu:
Everything needed to train a Qwen-GQA DSpark drafter for DeepSeek-V4-Flash-0731 from online disaggregated capture:

- **Official DeepSeek-V4 Python encoder**: the checkpoint ships a dedicated Python encoder rather than tokenizer Jinja. `DeepSeekV4Parser` invokes a vendored copy pinned to the model release revision, preserves the encoder's reasoning/tool fields, adapts dataset-level OpenAI tools to the encoder's system-message contract, and does not silently fall back to generic rendering when encoding fails.
- **v0.5.18 spec-capture patch — DSV4 additions** (applies cleanly to sglang `v0.5.18`): mHC prenorm prewarm fallback for `--chunked-prefill-size -1`; `last_hidden_states` channel + explicit-aux-wins-over-prenorm in `LogitsProcessor` (V4 always passes `hidden_states_before_norm`, which otherwise silently replaces the aux concat); capture-only `--spec-capture-method dspark` routing to the target's native `set_dspark_layers_to_capture` hook; a B200 `flashinfer_mxfp4` MoE fix.
- **`configs/deepseek-v4-flash-dspark.json`**: five-layer Qwen3-style GQA drafter at the target's width.
- **`examples/configs/online/disaggregated/external/deepseek-v4-flash-dspark-disaggregated.yaml`** + [runbook](https://github.com/sgl-project/SpecForge/blob/kan/pr2-dsv4-dspark/docs/recipes/deepseek-v4-flash-dspark-disaggregated.md): single-node recipe — two TP2 capture servers plus a four-rank trainer, from scratch on ShareGPT.

Validation:

- The vendored encoder body matches DeepSeek's release-revision `encoding_dsv4.py` byte-for-byte; only the retained MIT notice and pinned-source header are prepended.
- Basic chat, multi-turn thinking/drop-thinking, generation prompts, DSML tool rendering, parser selection, assistant-token loss masks, and fail-fast behavior are covered by focused tests.
- `35 passed, 41 subtests passed` across the focused DeepSeek-V4, data-parser, prompt-builder, collator, and disaggregated-launcher tests.
- Pre-commit and the GitHub lint workflow pass.
- The revised v0.5.18 patch passes `git apply --check` against the installed SGLang 0.5.18 source files.


specforge/data/template.py
```diff
@@ -313,6 +316,40 @@ def get_all_template_names(self) -> List[str]:
     ),
 )
 
+# DeepSeek-V4 checkpoints ship no Jinja chat template (prompts come from a
+# reference python encoder), so this template carries its own Jinja mirroring it.
+TEMPLATE_REGISTRY.register(
```
maocheng23, line 316: This looks not correct to me, can you please double confirm? especially the reasoning part

patches/sglang/v0.5.18/spec-capture.patch
maocheng23: This patch still lacks the `is_dspark` override that the K3 variant carries (`kimi-k3-f8493a4/spec-capture.patch` L718-726). Without it, `--spec-capture-method dspark` on `DeepseekV4ForCausalLM` falls through to the `ValueError` in stock `attention_backend_setup.py:63`, because V4 has no `set_dflash_layers_to_capture`. See the full chain and the verified hunk in my comment on the runbook (`docs/recipes/deepseek-v4-flash-dspark-disaggregated.md` L13-17). Once added, this comment line ("served by the kimi-k3 patch variant") should be rewritten.

specforge/data/template.py
jiapingW: Fixed in bd06d9e. I replaced the concatenated inline Jinja with a named, structured template and made its supported scope explicit. Chat mode now suppresses reasoning_content; thinking mode keeps reasoning only for the assistant turn after the final user, matching the reference encoder's default drop_thinking behavior; generation prompts select <think> vs </think> correctly. Unsupported tool/agent inputs fail loudly. Basic chat, multi-turn thinking, and both generation modes were compared byte-for-byte with DeepSeek's official encoding_dsv4.py and covered by regression tests.

patches/sglang/v0.5.18/spec-capture.patch
jiapingW: Fixed in bd06d9e. The v0.5.18 patch now passes capture-only --spec-capture-method dspark into configure_aux_hidden_state_capture by OR-ing the server capture mode with self.spec_algorithm.is_dspark(). This selects DeepseekV4ForCausalLM.set_dspark_layers_to_capture instead of falling through to the DFlash hook/error path. I also rewrote the misleading comment, added a patch-contract regression test, and verified the revised patch with git apply --check against SGLang 0.5.18.

PR #800, opened 2026-09-01
DSparkV4DraftModel: official DeepSeek-V4 DSpark drafter architecture

sherlockwu:
Stacked on #799 (retarget as the stack merges).

Adds training support for the official DeepSeek-V4 DSpark drafter architecture, so SpecForge can continue-train (or retrain from target weights) the drafter that ships with `deepseek-ai/DeepSeek-V4-Flash-0731`.

## What's here

- **`DSparkV4DraftModel`** (`specforge/modeling/draft/dspark_v4.py`): three DeepSeek-V4 blocks — MLA-style low-rank attention with grouped low-rank output projection, the 256-expert MoE (reusing `draft/moe.py` from #799), mHC hyper-connections with Sinkhorn splits, fp8-ue8m0 fake-quant on the KV path, Markov + confidence heads — under the official `mtp.*` checkpoint namespace. Module names mirror the official tensor names, so a trained state dict round-trips into the bundled layout without renaming.
- **`training.attention_backend: native`**: this drafter implements its own block/window attention pattern; the training wrapper skips mask construction and calls the draft model directly. The DSpark objective can feed the confidence head its own feature stream (`pop_confidence_hidden`) since V4's confidence head reads different features than the LM head.
- **Draft loading** validates against a provider's `compatible_architectures` (DSpark accepts both `DSparkDraftModel` and `DSparkV4DraftModel`).
- **Scripts**: `dspark_v4_official_weights.py` (download/inspect the official drafter), `init_dspark_v4_from_target.py` (initialize stages from target layers; identity-last or averaged `main_proj`; `--stage-layers` when capture layers differ from stage sources), `bundle_dspark_v4_official.py` (export a trained checkpoint into the official servable bundle), `gates/dspark_v4_parity_check.py` (numeric parity vs the reference implementation, tolerating single routing-flip outliers), `gates/dspark_v4_step_bench.py`, `bench_dspark_compute.py`.
- **Recipes**: official-arch two-node H200 (+ runbook) and 0813-traffic 32K on H200 and B200 2-node stacks, with launch scripts.

## Tests
- `tests/test_modeling/test_dspark_v4.py`: official state-dict naming and round-trips (including `from_pretrained`), forward/backward, context isolation, block independence, per-stage gradient-checkpointing equivalence, gate-bias fp32/deferred-update/eval-frozen semantics, fake-quant and Sinkhorn properties, and the `native`-backend end-to-end integration.
- Provider parity matrix and package gates updated (dspark gains the `native` backend; launch scripts registered as canonical CLI entries).

GPU validation on the training cluster: the parity gate passes against the official weights, and the two-node recipes are the ones behind the production official-arch runs.


fg11991:
### `MoEGate` sqrtsoftplus: NaN gradients from underflowed router logits

`specforge/modeling/draft/moe.py` (introduced in #799, used here by every
`DSparkV4Stage.ffn`):

```python
scores = F.softplus(scores).sqrt()
```

`softplus` underflows to **exactly** 0.0 in fp32 once a logit drops below about
-104. `sqrt` has an unbounded derivative there, so the backward evaluates
`grad / (2 * sqrt(0))` and produces NaN — even where the incoming gradient is
zero, which is every unselected expert. With 256 routed experts and a router
that is still moving, logits in that range are routine rather than exotic.

The failure signature is nasty: the **forward is finite and correct** (the NaN
never reaches the top-k weights, which come from the selected experts), and only
the gradient is poisoned. It looks like a data or an arithmetic problem
somewhere else entirely.

Reproduction against this branch's `MoEGate`, no GPU needed:

```python
import types, torch
from specforge.modeling.draft.moe import MoEGate

cfg = types.SimpleNamespace(
    scoring_func="sqrtsoftplus", num_experts_per_tok=6,
    routed_scaling_factor=1.5, n_routed_experts=256, hidden_size=64,
)
gate = MoEGate(cfg)
with torch.no_grad():                    # drive most logits into underflow
    gate.weight.zero_()
    gate.weight[:, 0] = -200.0
    gate.weight[:6, 0] = 0.5
x = torch.randn(4, 64); x[:, 0] = 1.0; x.requires_grad_(True)

w, idx = gate(x)
w.sum().backward()
print(torch.isfinite(w).all().item())            # True  <- forward is clean
print(torch.isnan(gate.weight.grad).sum().item())  # 16000 of 16384
print(torch.isnan(x.grad).sum().item())            # 256
```

Suggested fix — floor the softplus before the sqrt:

```python
-        scores = F.softplus(scores).sqrt()
+        # sqrt's derivative is unbounded at 0 and softplus underflows to
+        # exactly 0 below a logit of about -104 in fp32, which is routine
+        # with 256 experts; the backward then evaluates grad / (2*sqrt(0)).
+        scores = F.softplus(scores).clamp_min(1e-12).sqrt()
```

The floor is free: over 1000x256 ordinary logits the scores are **bitwise
identical** with and without it (max abs diff 0.0). `clamp_min` routes no
gradient to the floored entries, and 1e-6 against the ~1 of a selected expert
can change neither the top-k nor the normalized weights. The derivative at the
floor is `1 / (2 * sqrt(1e-12)) = 5e5` — large but finite.

The same line also fixes a second hazard two lines down:

```python
weights = weights / weights.sum(dim=-1, keepdim=True)
```

If every selected score underflows, that is `0 / 0`. With the floor in place the
degenerate case decays to a uniform `1/top_k` instead of NaN. (If you prefer to
be explicit, `.clamp_min(1e-20)` on the denominator is the other half.)

Happy to open this as a PR against the MoE branch if that is easier.

fg11991:
### `DSparkV4DraftModel` does not own the drafter's embedding and LM head

`dspark_v4.py` defines no `mtp.0.embed` / `mtp.<last>.head`, so the objective
falls through to the DFlash-family convention in
`specforge/algorithms/common/dflash_family_model.py`:

- `self.embed_tokens = target_embed_tokens` (L242), used by
  `_create_noise_embed` (L349) — the noise tokens handed to the drafter are
  embedded with the **target's** embedding;
- `self.lm_head = target_lm_head` (L241), used at L1179 for the draft
  `base_logits` and at L1210 for the teacher `target_logits` — both heads are
  the target's.

That convention holds for a drafter trained from scratch, and I think it is fine
for the GQA arm in #798. It does not hold for the official DSpark drafter: the
released checkpoint carries `mtp.0.embed.weight` and `mtp.2.head.weight`, and
neither matches the target's `embed.weight` / `head.weight`. On the
DeepSeek-V4-Flash-0731 weights I measured **max abs diff 3.71 and 4.44**, both
tensors stored unquantised — these are different tensors, not quantisation noise.

Warm-starting from the official weights and then scoring with the target's head
puts the drafter in a space it was never trained in. At step 1, against
`ln(129280) = 11.77` as the uniform baseline:

|                 | target head | drafter's own |
|-----------------|-------------|---------------|
| accuracy        | 0.025       | 0.584         |
| cross-entropy   | 9.267       | 2.181         |
| draft top-1 prob| 0.198       | 0.740         |

A from-scratch run cannot see this, which is why it survives to here.

Worth noting that configuration alone cannot fix it: `lm_head_key` feeds both
the draft objective head and the teacher head, so pointing it at
`mtp.2.head.weight` fixes one and breaks the other. What worked for me was
letting the stages own the two tables (loaded frozen, under the official
`mtp.*` names) and routing only the draft objective through them, while the
teacher distribution keeps using the target's head.

Two knock-on points:

1. `bundle_dspark_v4_official.py` asserts `expected_keys - set(state)` is empty
   against the official headers. If the official snapshot you bundle from
   carries `mtp.0.embed.weight` / `mtp.2.head.weight`, that assert should
   already be firing today — which suggests the bundler has not been run
   against a real snapshot end to end yet.
2. `dspark_v4_official_weights.py:21` notes that the `mtp.*` bundled with
   `DeepSeek-V4-Flash-0731` and the standalone `DeepSeek-V4-Flash-DSpark`
   release are different models. My measurement is on the Flash-0731 lineage
   (via its ModelSlim W8A8 republication, where both tensors are stored
   unquantised). If you are targeting the other release it is worth confirming
   the key set there before acting on this.

I have this working on my side and am glad to send it as a patch onto this
branch if useful.

PR #788, opened 2026-08-29
add dflash2 selector metrics

curnane-lab:
## Motivation

Follow-up to #772 (DFlash2). The DFlash2 selector currently reports only `selector_loss` / `selector_accuracy` / `selector_coverage` / `selector_target_probability`, which is not enough to answer the two questions that come up when tuning the selector:

- How much score mass can reranking ever move - i.e., is `selector_top_k` too small, or already wasteful?
- How much accepted length could a perfect reranker buy, and how much does the realized greedy path actually buy today?

This PR ports the three validation signals that the merged DFlash2 implementation in vllm-project/speculators reports (`unary_candidate_target_mass_at_k`, `unary_top_k_oracle_accepted_length`, and the realized greedy self-conditioned path), adapted to SpecForge's hard-target online pipeline. Since SpecForge captures hidden states only, the teacher-side soft "target mass" becomes the draft's own softmax mass on the same strict unary top-k candidate set.

## Modifications

`specforge/algorithms/common/dflash_family_model.py`:

- `SelectorTerms` / `DFlashObjectiveTerms` gain six additive telemetry fields (top-k mass, oracle/path accepted lengths, path accuracy, and their block/position denominators), preserving the flat tensor-only NamedTuple contract of `checkpointed_chunk_reduce`.
- Inside the existing `no_grad` block of `_selector_chunk_terms`, the new metrics are computed:
  - `unary_topk_probability_mass` - the draft's softmax mass inside its own strict unary top-k, via two `logsumexp` reductions (no extra full-vocab materialization);
  - `unary_topk_oracle_accepted_length` - per-block accepted length under a perfect reranker: every supervised slot whose gold token sits in the unary top-k is credited. The verified anchor counts as one accepted token, and a block extends only while every earlier slot is supervised and covered;
  - `selector_path_accepted_length` / `selector_path_accuracy` - the realized greedy candidate path: each slot is re-scored with its own selected predecessor (starting from the verified anchor) instead of the gold one, mirroring the serving walk; per-position accuracy is conditioned on the earlier path being correct.
- `forward` exposes the four new metrics through the existing `ratio_metrics` (numerator, denominator) contract under the same `_selector_objective_enabled` gate, so they chunk-reduce and all-reduce exactly like the current selector metrics (zero denominators are clamped by the consumer).
- Loss terms, gradients, and checkpoint contents are unchanged; drafts without a candidate selector pay nothing.

`tests/test_modeling/test_dflash2.py`:

[...]

PR #810, opened 2026-09-02
feat: add structured DFlash2 training metrics

maocheng23:
## Summary

- keep throughput and timing telemetry under `perf/*` instead of rewriting it as `train/perf/*`
- expose explicit `lk_loss`/`kl_loss` and `expected_acceptance` aliases in the main training/evaluation paths
- add aggregate and per-position hard-label metrics for every DFlash-family draft (unary top-1, unary top-k recall and probability mass, target probability/expected acceptance) under `train/dflash/*`, plus DFlash2-only selector metrics (loss, conditional accuracy, serving-path accuracy) under `train/dflash2/selector/*`
- add two per-block accepted-length proxies ported from #788: the unary top-k oracle walk and the realized greedy serving-path walk
- add two smooth expected-accepted-length proxies over the same per-block chain `1 + sum_k prod_{j<=k} a_j`: the gold-token probability (`dflash2/hard_label/expected_accepted_length`, the D-PACE surrogate) and the full-vocabulary expected acceptance against the target head (`dflash2/teacher/expected_accepted_length`, the sampling-regime counterpart)
- add loss-composition telemetry: the per-position share of effective objective weight (`position_<k>/objective/loss_weight_share`) and the LK-lambda CE weight (`train/objective/lk_kl_weight`)
- add online teacher-alignment metrics for expected acceptance, unary top-1 agreement, unary top-k probability mass, and selector serving-path agreement
- carry the optional target final hidden state through online capture/collation while preserving offline-v1 compatibility
- give `pad_and_concatenate_features` an `optional_keys` contract (present in every sample or none, mixed batches rejected) and collapse the DFlash/DSpark/MTP collators into one factory over it
- gate the full-vocabulary teacher/selector diagnostic pass to W&B logging steps and compute it chunked under `no_grad`

This complements the metric cleanup started in #797 and gives the DFlash2 dashboard a stable `perf/*`, `train/*`, `train/objective/*`, `train/dflash/{hard_label,teacher}/*`, `train/dflash2/selector/*`, and `position_<k>/{hard_label,objective,teacher,selector}/*` hierarchy.

## Credits

The accepted-length and candidate-mass metrics are ported from #788 by @curnane-lab, credited as co-author on the porting commit. This PR supersedes #788 for those metrics; the realized path is computed from the shared `greedy_path` walk instead of a separate scoring loop.

## Behavior notes

- Per-position families start at predicted position 1; the non-predicted anchor position is omitted. They are keyed `position_<k>/<family>/<metric>` so W&B renders one section per block position; aggregates stay under `train/dflash/*` and `train/dflash2/selector/*`.
- Top-K diagnostics use the selector's `selector_top_k` when present, else the model's `metric_top_k` (default 16, clamped to the vocabulary), so plain DFlash drafts get the same unary recall, mass, and oracle accepted-length readouts.
- Accepted-length proxies count the verified anchor as one token and extend a block only while every earlier slot is supervised and accepted; the denominator is the number of blocks with at least one supervised slot. The expected variants chain per-slot probabilities per block rather than multiplying per-position means, and all of them are teacher-forced on the real prefix and anchor.
- `selector_conditional_accuracy` is uniform over covered slots and therefore comparable between `dflash` and D-PACE runs. The historical `selector_accuracy` and `selector_loss` remain weighted by the effective objective weights.
- `objective/lk_kl_weight` is reduced as a token-weighted mean of the per-micro-batch CE weight.
- Teacher metrics are available for online capture data and are omitted when offline data does not contain target final hidden states.
- Ratio metrics are reduced by numerator/denominator across data-parallel ranks before logging.

## Validation

- `139 passed, 30 subtests passed` in the focused CPU suite (`tests/test_modeling/test_dflash2.py`, `tests/test_runtime/test_tracking_logger.py`, `tests/test_utils/test_dflash_losses.py`, `tests/test_runtime/test_trainer.py`, `tests/test_algorithms/test_builtin_parity.py`, `tests/test_runtime/test_seam_fixes.py`, `tests/test_runtime/test_evaluator_aggregation.py`)
- multi-process evaluator reduction test passed
- `black --check` passed for all changed Python files
- `git diff --check` passed
- the SGLang-backed server capture gate was not run locally because SGLang is not installed; its capture schema assertions were updated and the non-gate capture tests pass

## Follow-ups (not in this PR)

- Online runs still cannot run a held-out evaluation (`data.eval_data_path` is rejected for online mode and `eval_hidden_states_path` is offline-only), so every metric here is a training-set metric on the current micro-batch.
- W&B run config records the draft config path but not its content, and custom launch scripts should pass the git commit to the tracker.

### DIFF
--- a/docs/basic_usage/training.md
+++ b/docs/basic_usage/training.md
@@ -224,17 +224,22 @@ anchor is omitted), so each position renders as one dashboard group:
 - `hard_label/*`: unary top-1 accuracy, top-K recall and probability mass,
   gold-token probability, selector loss, selector conditional accuracy (uniform
   over covered slots, so comparable across loss types), and the realized greedy
-  serving path's per-slot accuracy. Two aggregate-only per-block accepted-length
-  proxies count the anchor plus the leading run of supervised slots that are
-  covered by the unary top-K (`unary_topK_oracle_accepted_length`) or matched by
-  the serving path (`selector_serving_accepted_length`).
+  serving path's per-slot accuracy. Three aggregate-only per-block
+  accepted-length proxies apply `1 + sum_k prod_{j<=k} a_j` over the supervised
+  slots with `a_j` being coverage by the unary top-K
+  (`unary_topK_oracle_accepted_length`), a serving-path hit
+  (`selector_serving_accepted_length`), or the gold-token probability
+  (`expected_accepted_length`, the smooth D-PACE surrogate). All three are
+  teacher-forced on the real prefix and anchor.
 - `position_<k>/objective/loss_weight_share`: the fraction of the effective
   objective weight each block position receives under the configured
   fixed-decay or D-PACE weighting.
 - `train/objective/lk_kl_weight`: the CE weight of the `lambda` LK objective.
 - `teacher/*` (online capture only): full-vocabulary expected acceptance, unary
   top-1 agreement, unary top-K teacher mass, and serving-path agreement against
-  the frozen target head.
+  the frozen target head, plus the aggregate `expected_accepted_length` that
+  chains the per-slot expected acceptance, the sampling-regime counterpart of
+  the hard-label surrogate.
 
 The exported computation and parameter names match the public SGLang DFlash2
 contract, including optional `output_multiplier` and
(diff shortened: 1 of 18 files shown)


FrankLeeeee (approved):
LGTM.

PR #823, opened 2026-09-03
feat(dflash): add full-vocabulary AngelSpec LK loss

maocheng23:
## Summary

- add an opt-in, full-vocabulary LK objective for online DFlash / DFlash2
- compute AngelSpec acceptance, TV, KL, and the LK-lambda coefficient independently per token
- compose that token loss with the existing detached DFlash or D-PACE position weights
- reuse the target final-hidden-state capture merged in #810
- keep the current hard-label objective as the default and fingerprint the supervision source in the resume contract

## Motivation

The existing DFlash LK path treats the observed token as a one-hot target. That makes LK-alpha equivalent to CE and derives LK-lambda from an acceptance rate aggregated over the batch. AngelSpec instead compares the full teacher and draft vocabulary distributions and computes lambda per token.

## Objective

For frozen teacher distribution `p` and trainable draft distribution `q`:

```
TV_i       = 0.5 * sum_v |p_i(v) - q_i(v)|
accept_i   = 1 - TV_i
lambda_i   = kl_scale * exp(-kl_decay * stop_grad(accept_i))
LK_i       = lambda_i * KL(p_i || q_i) + (1 - lambda_i) * TV_i
```

`alpha` uses `-log(accept_i)`; `tv` uses `TV_i`. The configured DFlash/D-PACE position weight is then applied to `LK_i`.

This follows the [AngelSpec DFlash objective](https://github.com/Tencent/AngelSpec/blob/main/angelspec/models/dflash.py#L452-L476) and its [LK loss helper](https://github.com/Tencent/AngelSpec/blob/main/angelspec/models/ops/loss.py#L220-L255).

## Configuration

```yaml
training:
  strategy: dflash
  loss_type: dpace
  dpace_alpha: 0.5
  lk_loss_type: lambda
  dflash_lk_target: target_distribution
  kl_scale: 1.0
  kl_decay: 3.0
```

`dflash_lk_target` defaults to `hard_label`, so existing recipes and checkpoints keep their current behavior. `target_distribution` is currently online-only because offline DFlash features do not store `target_last_hidden_states`; schema and runtime checks fail explicitly if the capture contract is not satisfied.

## Validation

- pre-commit checks on all changed files
- 137 focused CPU tests across LK/DFlash/DFlash2/Domino, schema, provider, and strategy suites (1 CUDA-only test skipped)
- all 57 tests under `tests/test_algorithms`
- formula parity, frozen-teacher gradients, D-PACE composition, missing-feature failure, and chunked/non-chunked loss + gradient parity

Draft because this has not yet been exercised in a GPU optimizer run.


maocheng23:
Status note (2026-09-08): in the Qwen3.8-27B DFlash2 loss ablation (RadixArk internal, one epoch, 512 samples/step, AIME25/26 acceptance length paired on identical prompts), the full-vocabulary teacher-distribution LK arm (T1: lk_loss_type lambda, dflash_lk_target target_distribution, eta 3) came out a null result against the hard-CE baseline: +0.022 acceptance length, 95% CI [-0.035, +0.079]. The selector objective (strict top-k) was the only loss lever that moved serving (+0.14). Keeping this PR as a draft for a later multi-epoch stage-2 test rather than merging it now; it is not a dependency of the KDA work in #836.

PR #832, opened 2026-09-03
fix(mooncake): treat OBJECT_NOT_FOUND as a completed removal

maocheng23:
## Problem

`MooncakeFeatureStore._store_remove` treats every non-zero return code as a failed removal. A sample's tensors are freed one key at a time, so the optimizer-boundary `abort()` (no force) removes the keys whose read lease has expired and gets `OBJECT_HAS_LEASE` (-706) on the rest; the sample stays in `_release_pending`. The forced retry at the next boundary (`retry_sample_removals`, no probing) then gets `OBJECT_NOT_FOUND` (-704) for the keys that are already gone and counts that as a failure too. The sample can only be classified as freed by the escalation drain (`_CLEANUP_ESCALATION_BOUNDARIES`), whose 40 x 0.25 s retry loop probes `is_exist` only on the final attempt.

Observed on a Qwen3.8-27B NVFP4 / DFlash2 disaggregated run (1 capture server + DP7 trainers, B300, sglang 0.5.18, mooncake-transfer-engine-cuda13 0.3.13): the durable ack alternated between 0.01 s and exactly 9.78 s on every rank, 50% of rank time, independent of the lease TTL (3 s and 5 s both).

## Verified Mooncake semantics (standalone `mooncake_master`, 0.3.13 binding)

| call | rc |
|---|---|
| `remove(key)` during a live read lease | -706 |
| `remove(key, force=True)` during a live read lease | 0 |
| `remove(key)` / `remove(key, force=True)` on an absent key | -704 |
| `is_exist(key)` | renews the lease (`remove` right after it fails with -706) |

## Change

`_store_remove` returns True for rc 0 **or -704**; the two codes get named constants. No other behaviour changes: leased keys still fail with -706 and go through the existing retry/force path.

## Test

`test_retry_treats_already_removed_keys_as_freed`: a fake store whose first free removes the unleased key and fails the leased one with -706, then answers -704 for the removed key on the forced retry. Before the fix the sample stays pending; after it `retry_sample_removals` frees it. `tests.test_runtime.test_mooncake_store`, `test_offline_mooncake_cleanup`, `test_mooncake_architecture` pass.

### DIFF
--- a/specforge/runtime/data_plane/mooncake_store.py
+++ b/specforge/runtime/data_plane/mooncake_store.py
@@ -80,6 +80,10 @@ from specforge.runtime.data_plane.feature_store import (
 logger = logging.getLogger(__name__)
 
 # Defaults for MooncakeDistributedStore.setup(); override via ``setup_kwargs``.
+#: Mooncake ``ErrorCode`` values the store reasons about explicitly.
+MOONCAKE_OBJECT_NOT_FOUND = -704  # remove()/get() on an already-freed key
+MOONCAKE_OBJECT_HAS_LEASE = -706  # remove() during a live read lease
+
 _MOONCAKE_SETUP_DEFAULTS = {
     "global_segment_size": 1 << 30,  # 1 GiB per-node segment
     "local_buffer_size": 1 << 30,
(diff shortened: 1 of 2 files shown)


jiapingW (approved):
LGTM

PR #835, opened 2026-09-04
feat(dflash): report sampled serving-walk accepted length

lvhua898-cpu:
## Motivation

DFlash-family training previously averaged accepted length over every independently sampled anchor. Serving does not visit every anchor: after verifying one block, it resumes from the first sampled anchor at or after the next required token. The per-block average can therefore diverge from the acceptance length of the path that serving would actually take.

The prediction path also differs across the family. DFlash2 uses its autoregressive candidate selector, Domino conditions its correction head on already generated tokens, and DSpark autoregressively applies its Markov head. A useful training-time serving proxy must replay those paths instead of using teacher-forced logits everywhere.

## Modifications

- Add a strict accepted-prefix helper: the prefix stops at the first incorrect or unsupervised token, so a masked hole cannot be skipped.
- Add a sampled-anchor walk reducer that advances by token position and reconstructs global anchor order before walking, making the result independent of `objective_chunk_blocks`.
- Report the following additive numerator/denominator metrics through the trainer's distributed `ratio_metrics` path:
  - `dflash/hard_label/walk_accepted_length`
  - `dflash2/selector/walk_accepted_length`
  - `domino/final/walk_accepted_length`
  - `domino/base/walk_accepted_length`
  - `dspark/hard_label/walk_accepted_length`
- Use each family's serving prediction semantics:
  - DFlash: unary hard-label argmax.
  - DFlash2: `candidate_selector.greedy_path`, including its autoregressive predecessor transitions.
  - Domino: greedily replay the token-conditioned correction head; support both shifted and unshifted label layouts and exclude only the structural anchor/training-only slot from acceptance.
  - DSpark: greedily replay `sample_block_tokens` when a Markov head is present, seeded by the anchor token; retain unary argmax as the no-Markov fallback. The full next-token block matches SGLang's `gamma`-token DSpark serving layout.
- Fix unshifted Domino acceptance, where the deliberately masked anchor slot previously forced every strict accepted prefix to zero.
- Run all new serving-path collection only when `collect_detailed_metrics=True`; ordinary non-log steps keep the existing objective path and avoid the autoregressive telemetry work.
- Preserve the latest `main` behavior that exposes Domino `loss_terms` for globally normalized distributed loss, while independently forwarding the new walk ratios.

These metrics are diagnostic only and do not change the loss or backward graph. They replay the path over sampled training anchors, so sparse anchor sampling remains a proxy rather than an exact end-to-end SGLang evaluation.

## Tests

- Deterministic numerator/denominator checks for unary, selector, Domino final/base, and DSpark Markov walks.
- Shifted and unshifted Domino label-layout coverage.
- Strict-prefix regressions for masked holes and empty/non-prefix valid-anchor masks.
- Chunk-invariance checks with `objective_chunk_blocks` values `0`, `1`, `2`, `4`, and `5`.
- Detailed-metrics gating checks that verify Domino/DSpark autoregressive serving helpers are not called on ordinary steps.
- DFlash2, Domino, DSpark model tests plus Domino/DSpark offline and disaggregated FSDP runtime tests.
- `pre-commit run --files specforge/algorithms/common/dflash_family_model.py specforge/training/strategies/base.py tests/test_utils/test_dflash_losses.py` — passed.
- `pytest -q tests/test_utils/test_dflash_losses.py tests/test_modeling/test_domino_model.py tests/test_modeling/test_dflash2.py tests/test_modeling/test_draft_registry.py tests/test_runtime/test_domino_launch.py tests/test_runtime/test_dspark_disagg_launch.py tests/test_runtime/test_dspark_offline_launch.py tests/test_runtime/test_seam_fixes.py` — 107 passed, 28 subtests passed.

## Performance

No additional serving-walk work is performed when `collect_detailed_metrics=False`. Detailed-metrics steps reconstruct small `[batch, sampled_anchors]` tensors and execute the sequential sampled-anchor walk on CPU. DFlash2, Domino, and DSpark additionally replay their autoregressive prediction heads on those steps; the cost therefore follows `training.log_interval` and does not affect non-log optimizer steps.


jiapingW:
help cc. @maocheng23

maocheng23:
Closed as #880 merged

PR #840, opened 2026-09-06
feat(mooncake): pooled pinned or device receive buffers for disaggregated feature reads

maocheng23:
## Summary

Disaggregated feature reads currently allocate and register a fresh pageable host buffer for every tensor. This adds opt-in `pinned` and `cuda` receive pools, retaining registered slots across reads and returning caller-owned tensors to the loader. The default remains `pageable`.

- `receive_buffers` and `receive_pool_bytes` are validated and forwarded to worker processes. Both pooled modes request the rank's CUDA device when available, so pinned H2D copies run in loader workers.
- H2D/D2D copies complete before a slot can be reused. Failed transfers retain their backing storage in quarantine, including overflow allocations; late DMA cannot target recycled memory.
- Failed host registration raises immediately without disabling future registrations or consuming the pool budget. CUDA registration failures use Mooncake's device-read staging path with one warning.
- Replica-not-ready, expired-read-lease, and transfer failures retry into fresh storage. Invalid parameters, exhausted client staging space, and short reads fail directly.
- Health counters report reuse, overflow, and quarantined bytes. The budget limits retained slots; concurrent overflow and returned tensors consume additional memory.

## Mooncake compatibility

Mooncake 0.3.x stages remote device-destination reads through its pinned client buffer before copying to CUDA. This is not NIC-direct receive into trainer memory. Size `deployment.disaggregated.client_buffer_size` for the largest simultaneous reads; a 272 MiB tensor already exceeds the default 256 MiB buffer. The H200 tests used 2 GiB.

PR #841 enables device publication on the server independently of this receive-side change.

## Validation

Validated on two H200 nodes with PyTorch 2.13.0+cu130, SGLang 0.5.18, Mooncake 0.3.13.post1, and InfiniBand:

- CPU regressions and CUDA tests for late-write quarantine, registration recovery, transient retries, copy completion, output ownership, and unindexed CUDA-device reuse.
- Combined #840/#841 regression suite: 173 tests on each node, successful with one existing skip and one expected failure.
- Real cross-node 272 MiB BF16 tensors: RDMA host/GPU publication to pageable/pinned/CUDA receivers; TCP host publication to pageable/pinned receivers. Exact equality under serial and four-worker reads, retained-slot reuse, overflow, and two-worker FeatureDataLoader prefetch.
- Actual patched SGLang capture across nodes: 48 samples across 12 publication/algorithm/receive combinations, independent Hugging Face hidden-state/logit parity, CPU/GPU publication equality, and four finite-loss Eagle3 optimizer steps.
- `pre-commit run --all-files` passed.

These are correctness regressions, not a rerun of the historical Qwen/B300/H200 throughput study. Earlier performance numbers have been removed because they were not established for these exact PR revisions.


jiapingW:
@maocheng23 Can you help veirfy the lastest impl in 2xH200?

jiapingW (approved):

PR #841, opened 2026-09-06
feat(spec-capture): publish captured tensors from device memory (GPU put)

maocheng23:
## Summary

Enabling GPU publication at the sink alone leaves captured tensors on the CPU because the scheduler stages them before submission. This change keeps captured features on CUDA throughout the scheduler-to-Mooncake path when `SGLANG_SPEC_CAPTURE_GPU_PUT=1` is enabled.

- Pure capture batches skip hidden-state D2H; ordinary hidden-state responses remain supported when requests share the server.
- The writer rank snapshots batch tensors before their storage can be reused. CUDA stream dependencies prevent the next forward from overwriting graph outputs before the snapshot completes, and readiness events hand snapshots safely to the asynchronous writer.
- Per-request views register their shared underlying storage once. Successful registrations are unregistered after synchronous publication; GPU registration failure stops publication before returning refs.
- Timing logs include GPU/host payload bytes, making the selected path observable.
- Managed-local `capture_servers[].gpu_put` renders the flag and requires RDMA. Externally configured GPU publication also rejects TCP. CUDA allocations must be RDMA-registerable; documentation covers PyTorch expandable-segment compatibility.

The flag defaults to false and the feature wire format remains unchanged. Device snapshots occupy GPU memory until their queued store writes complete. This is independent of the receiver pooling in #840; Mooncake 0.3.x still stages remote CUDA receives through its client buffer.

## Validation

Validated with #840 on two H200 nodes using PyTorch 2.13.0+cu130, SGLang 0.5.18, Mooncake 0.3.13.post1, and InfiniBand:

- CPU/CUDA regressions cover shared-storage registration, premature CPU staging, non-default-stream handoff, source overwrite, forward-stream ordering, and failed GPU registration.
- Combined regression suite: 173 tests on each node, successful with one existing skip and one expected failure.
- Real 272 MiB GPU-source publication, followed by exact cross-node reads through pageable/pinned/CUDA receivers under serial and four-worker load. Logs confirm 285,212,672 GPU payload bytes per synthetic sample.
- Live patched SGLang: CPU/GPU publication × Eagle3/DFlash × three receive modes, 48 samples total. Hidden states and recomputed logits match an independent Hugging Face reference at the fixture's BF16 tolerance; CPU/GPU publication tensors match exactly. Four Eagle3 optimizer steps produce finite losses. Concurrent ordinary and capture requests both succeed.
- The complete patch applies to its exact upstream source blobs and all 12 patched modules parse. `pre-commit run --all-files` passed.

These tests use a deterministic tiny Llama target plus representative large transport payloads; they do not establish large-model training throughput. The previous benchmark claims were removed because the original PR's scheduler still performed CPU staging.


jiapingW (approved):
LGTM.

PR #872, opened 2026-09-09
fix(export): honor explicit vocabulary mappings in HF exports

StevenWang-CY:
When a checkpoint already contains `t2d`/`d2t`, `specforge export --to hf --vocab-mapping replacement.pt` loads the requested mapping, then silently overwrites it with the checkpoint buffers before saving. The resulting draft decodes output indices using the old target-token IDs. The SGLang exporter honors the same override.

Preserve the explicitly loaded mapping when restoring checkpoint tensors for serialization. All other checkpoint tensors keep their original values and precision. Exports without an override and legacy checkpoints missing mapping buffers retain their existing behavior.

Four CPU regression tests inspect saved safetensors and reload the actual draft with no missing/unexpected keys. They cover an existing mapping override, a legacy checkpoint, no override, and the real CLI with an external frozen embedding. Non-mapping weights are checked bit-for-bit, including FP32 values that BF16 materialization would round. Loading-info assertions require empty key collections and accept the lists returned by Transformers 4 or the sets returned by Transformers 5.

Validation on `a4dd2ff663a31bab3e7a82ebe6620a33a4e39fbf`:

- [Hosted lint](https://github.com/sgl-project/SpecForge/actions/runs/34407339291): passed.
- [Hosted PR Test](https://github.com/sgl-project/SpecForge/actions/runs/34407339298): both live CUDA server-capture tests passed; full discovery completed successfully with 1,173 tests, 10 skips and 1 expected failure. The checkout was the merge of this exact head with main `fcc0c36750dce1c9b505d83ee318bd70134c2a46`.
- Fresh local checkout: 86 focused/neighbor tests completed successfully under each of Transformers 4.57.6 and 5.12.1 (7 skips per run). Full `pre-commit run --all-files` passed.
- The two mapping-override regressions still fail against unmodified main with Transformers 5.12.1, while the legacy and no-override controls pass. All four pass on this head.

The previous CI failure was reproduced locally with Transformers 5.12.1: the test compared an empty set of missing keys to an empty list. Checking collection emptiness fixes that assertion without relaxing the missing/unexpected-key or tensor-equality checks. The production exporter change is unchanged.

### DIFF
--- a/specforge/export/to_hf.py
+++ b/specforge/export/to_hf.py
@@ -112,7 +112,15 @@ def export_to_hf(
         full_state["embed_tokens.weight"] = _load_embedding_tensor(
             embedding_source, embedding_key
         )
-    full_state.update(state["draft_state_dict"])  # trained keys win
+    # Preserve checkpoint weights at their original precision, but keep the
+    # mapping explicitly loaded by materialize_draft when one was requested.
+    full_state.update(
+        {
+            key: value
+            for key, value in state["draft_state_dict"].items()
+            if not (vocab_mapping_path and key in {"t2d", "d2t"})
+        }
+    )
     model.save_pretrained(output_dir, state_dict=full_state)
     apply_legacy_rope_scaling(output_dir)
     return output_dir
--- a/tests/test_runtime/test_export.py
+++ b/tests/test_runtime/test_export.py
@@ -165,6 +165,143 @@ class TestLegacyVocabMappingCompatibility(unittest.TestCase):
             )
 
 
+class TestHFVocabMappingExport(unittest.TestCase):
+    """Exercise serialized and reloaded exports on CPU, without model downloads."""
+
+    def setUp(self):
+        from specforge.modeling.auto import AutoDraftModel, AutoDraftModelConfig
+        from tests.test_runtime import _fixtures as fx
+
+        self.tempdir = tempfile.TemporaryDirectory(prefix="hf_mapping_export_")
+        self.addCleanup(self.tempdir.cleanup)
+        self.cfg_path = fx.write_draft_config(
+            os.path.join(self.tempdir.name, "draft.json")
+        )
+        old_path = fx.write_vocab_mapping(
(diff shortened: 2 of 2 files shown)


StevenWang-CY:
@jiapingW, could you review this HF exporter fix? You merged #742 in this area, so your guidance on mapping precedence would be helpful.

A checkpoint containing `t2d`/`d2t` silently overrides the mapping supplied through `--to hf --vocab-mapping` when the exporter saves the weights. The same input to the SGLang exporter honors the replacement. This change preserves the requested mapping while keeping all other checkpoint tensors at their original precision.

The two override regressions fail on clean main and pass here; the four new CPU tests inspect saved tensors and reload the real model, including through the CLI. The PR description includes focused tests, separate export checks, and full-suite comparisons against main. Hosted lint and GPU tests are running. If someone else is better placed to review this path, I'd appreciate a referral.

StevenWang-CY:
@maocheng23, could you review this HF export fix? You authored the HF/SGLang exporter path, so your input on mapping precedence would be helpful.

When a checkpoint already contains `t2d`/`d2t`, `--to hf --vocab-mapping ...` loads the requested mapping, then silently overwrites it with the old checkpoint buffers during serialization. The SGLang exporter honors the override. This fix preserves the explicit mapping while retaining every other checkpoint tensor's original values and precision.

Validation on head `a4dd2ff`:

- All four regression tests pass, covering saved tensors, real-model reload, CLI/external embedding, legacy checkpoints, and no override. Rechecked against main `3d64e7a`: both override cases still fail there, while the two compatibility controls pass.
- [Lint](https://github.com/sgl-project/SpecForge/actions/runs/34407339291) and [PR Test](https://github.com/sgl-project/SpecForge/actions/runs/34407339298) passed, including two live CUDA capture tests and the 1,173-test suite. Hosted CI tested this head merged with main `fcc0c36`.

The main review points are whether the explicit override takes precedence at serialization and whether the legacy/no-override behavior remains correct. Thanks!

maocheng23:
LGTM

PR #876, opened 2026-09-12
offline capture: compat shims for sglang-main API drift (sglang 0.5.18 path unchanged)

bighead-liat:
Follow-up to #870, as requested there: the sglang-main API drift the offline capture backend hits, handled as shims that keep the pinned `sglang==0.5.18` path unchanged — not patches.

## What drifted (sglang main, Sep 2026)

| call in `sglang_backend/capture.py` | 0.5.18 | sglang main |
|---|---|---|
| `sglang.srt.utils.require_mlp_sync` / `require_mlp_tp_gather` | take `ServerArgs` | take no arguments (read the runtime context) |
| `ServerArgs.device` | resolved at construction | stays `None` until the launcher's post-process pass; `ModelRunner` needs a concrete device |
| `sglang.srt.runtime_context.publish(...)` | exists, not required | must run before a `ModelRunner` is built |
| KV allocator for hybrid Mamba/GDN/KDA targets | — | `page_size=1` lands on an allocator without `alloc_extend`; needs a paged allocator |

## Changes

- `specforge/offline_capture/sglang_backend/compat.py` (new): `require_mlp_sync` / `require_mlp_tp_gather` dispatch on the installed signature; `resolve_device` fills `device` only when it is `None`; `publish_runtime_context` publishes once and is a no-op when already published or when the API is absent. On 0.5.18 each helper reduces to exactly the previous call.
- `capture.py`: imports those four names from `.compat` instead of `sglang.srt.utils` and calls `resolve_device` + `publish_runtime_context` before `ModelConfig` / `SGLangRunner`. It still owns the direct SGLang imports (the boundary test is unchanged).
- `scripts/prepare_hidden_states.py`: `--sglang-page-size` and `--sglang-moe-runner-backend`, forwarded to `ServerArgs` only when set, so the pinned defaults stay in force otherwise. The page size is what hybrid targets on recent SGLang need (`64` worked for Ling-3.0-tiny); the MoE backend is for hosts whose CUDA toolkit cannot build the flashinfer fused-routing kernels.
- `tests/test_runtime/test_sglang_main_compat.py`: both signatures for the two predicates, device fill-in only when missing, single publish / no-op when already published or absent, and an AST check that `capture.py` routes the drifted calls through `compat` and runs the two setup shims before `SGLangRunner`.

## Verified

- Unit tests: `test_sglang_main_compat`, `test_sglang_0518_compat`, `test_sglang_capture_backend` — 14/14 on sglang 0.5.18 and 14/14 on sglang main (`de0fa171`).
- Real `ServerArgs` on both versions (CPU): `resolve_device`, publish twice (True then False), `publish_role() == "scheduler"`, both predicates return.
- Offline capture end to end on sglang main with Ling-3.0-tiny (`--strategy dspark`, `--sglang-page-size 64 --sglang-moe-runner-backend triton --sglang-attention-backend triton`, 16 rows) with **no local edits** — this is the run that previously needed hand-applied shims (#870).
- Offline capture on 0.5.18 could not be exercised on my host: its JIT kernels (`qknorm`, `store_cache`) fail with `CUDA error: invalid resource handle` there regardless of this PR (CUDA-11 system toolkit). The 0.5.18 code path is covered by the unit tests above and by the existing CI unit-test job.

### DIFF
--- a/scripts/prepare_hidden_states.py
+++ b/scripts/prepare_hidden_states.py
@@ -193,6 +193,25 @@ def parse_args():
     sglang_group.add_argument("--sglang-enable-dp-lm-head", action="store_true")
     sglang_group.add_argument("--sglang-ep-size", type=int, default=1)
     sglang_group.add_argument(
+        "--sglang-page-size",
+        type=int,
+        default=None,
+        help=(
+            "KV-cache page size for the offline SGLang capture (default: SGLang's "
+            "default). Hybrid Mamba/GDN/KDA targets on recent SGLang need a paged "
+            "allocator, e.g. 64."
+        ),
+    )
+    sglang_group.add_argument(
+        "--sglang-moe-runner-backend",
+        default=None,
+        help=(
+            "MoE runner backend for the offline SGLang capture (default: SGLang's "
+            "auto). Use e.g. triton on hosts whose CUDA toolkit cannot build the "
+            "flashinfer fused-routing kernels."
+        ),
+    )
+    sglang_group.add_argument(
         "--sglang-disable-radix-cache",
         action="store_true",
         help=(
@@ -294,7 +313,7 @@ def _generate_shared_vocab_mapping(
 
 
 def _sglang_kwargs(args: argparse.Namespace) -> Dict[str, object]:
-    return {
+    kwargs = {
         "attention_backend": args.sglang_attention_backend,
         "mem_fraction_static": args.sglang_mem_fraction_static,
         "context_length": args.sglang_context_length,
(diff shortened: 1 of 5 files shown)


bighead-liat:
Addressed both review points in 561185f: `resolve_device` now uses `specforge.utils.get_device_type()` (honours `SPECFORGE_DEVICE`, then CUDA, NPU, CPU) instead of probing CUDA alone, and `test_prepare_hidden_states` gained a focused test that the two optional engine knobs are forwarded as `page_size` / `moe_runner_backend` only when set. 28/28 on both sglang 0.5.18 and main.

PR #884, opened 2026-09-16
feat(moe): partition the routed experts across an expert-parallel group (moe.py variant)

fg11991:
Partitions the routed experts across an expert-parallel group, so a draft MoE
too large to replicate can still be trained. On the drafter we are training,
19.3B of 19.85B trainable parameters are routed experts, which do not fit
replicated on eight 64 GB cards.

Stacked on #799 for the MoE itself, and on #882 for the runtime half
(group topology, gradient-norm exactness, optimizer state, checkpointing). It is
inert without #882: `draft_ep_layout()` finds no group and the layer builds
exactly as it does today, so this PR is safe to land in either order. The tests
drive the seam directly rather than provisioning a mesh, so they stand alone.

The same change against the `moe/` package of #812 is in the companion PR, since the two MoE implementations are still both in flight — please take whichever matches the expert layer you intend to land, and close the other.

## Approach

The experts are three stacked `[E, out, in]` parameters, so sharding is a slice
along `E`: each rank allocates only its own experts and computes only the tokens
routed into them, while the tokens stay replicated. The combined output is the
group-wide sum, while autograd follows the local expert graph — expert gradients
stay rank-local and are never reduced twice.

Sorting the routing by expert already makes a rank's experts one contiguous run
of slots, so the local work is a slice of the sorted order. Reading the bounds
costs the host sync the sorted loop already pays; under `grouped_mm` it is the
price of expert parallelism.

**Where the seam goes.** It wraps the router and the experts, but not the shared
expert. Under EP both the router and the experts hold a *partial* `dL/dx` — the
router because only the slots of locally owned experts carry gradient, the
experts because this rank ran only its own — and the seam sums them. The shared
expert is replicated and already sees the whole loss; reducing it too would
multiply it by the group size. The router's *weight* gradient is partial for the
same reason and is all-reduced by a parameter hook.

**A rank that owns no selected expert.** Routine on an imbalanced route, and the
subtle case: that rank's input and routing weights then have no differentiable
consumer at all, so autograd never reaches the seam, its all-reduce is never
issued, and the peers that issued theirs hang. Zero-valued terms keep the input,
the combine weights and the expert parameters on the graph, so the graph — and
with it the collective order — is identical on every rank. This one cost me an
afternoon; the regression test for it fails (fast, not by hanging) if the terms
are removed.

**Checkpoints** keep global expert indices. `expert_offset` travels with the
stacked parameters so the converter can restore them and is dropped from the
file, which leaves a shard described by its indices alone. Loading slices a full
stacked tensor down to this rank, so warm start and resume do not care what
topology wrote the payload. The offset itself is never loaded — it describes
this rank, not the checkpoint.

## Tests

`tests/test_modeling/test_moe_expert_parallel.py`, two gloo ranks:

- a sharded layer is numerically the unsharded layer — same output, same input
  gradient, same router gradient, and each rank's expert gradients equal the
  matching slice of the reference's;
- a rank with no selected expert still reaches every collective (the starvation
  is forced through the selection-only balancing bias, so the scores are
  untouched — driving the gate *weight* negative instead pushes logits into
  softplus underflow and produces NaN gradients, which is a different problem);
- without a group the layer is unsharded and the checkpoint naming is unchanged.

Existing MoE tests pass unchanged, as do `tests/test_optimizer` and
`tests/test_config`. (Local runs need Python 3.11; the repo's `X | None`
annotations do not import under the 3.9 my system ships.)


fg11991:
Companion against the #812 `moe/` package: https://github.com/sgl-project/SpecForge/pull/883 — same change, different expert layer. Only one of the two should land.

fg11991:
The `lint` check had one failure of mine — isort on
`tests/test_modeling/test_moe_expert_parallel.py` — now fixed and pushed.

The remaining one is inherited, not from this PR: `pre-commit run --all-files`
lints the whole tree, and `black` reformats `specforge/modeling/draft/dflash.py`
on `kan/pr3-moe-drafter` with nothing of mine applied. Running the hooks over
only the three files this PR touches (`draft/expert_parallel.py`, `draft/moe.py`,
`tests/test_modeling/test_moe_expert_parallel.py`) passes clean.

Worth fixing at the source, since #799 will hit it on its own. I have kept it out
of this PR rather than touch a file that belongs to your in-flight work — say the
word and I will add it as a separate commit here to get the check green.

PR #894, opened 2026-09-19
Register glm-5.3 chat template

hiteshjain118:
## What

Registers a `glm-5.3` entry in `TEMPLATE_REGISTRY`, so GLM-5.3 draft-training configs can name their actual model family instead of borrowing `glm-5.2`.

GLM-5.3's chat template (zai-org) renders the same role/think tokens as GLM-5.2: `\u003c|user|\u003e` headers, the think opener emitted as part of the generation prompt right after the assistant header, and `\u003c|user|\u003e` as the turn separator — so the field values match the existing `glm-5.2` registration. The separate entry exists so any future 5.3-specific format drift can be handled without affecting `glm-5.2` users.

## Test

Adds `tests/test_data/test_glm_53_template.py` (mirrors `test_kimi_k3_template.py`): asserts the registration exists with the `glm` parser contract and that the glm-5.2 / glm-5.3 field sets stay in sync.

## Verification

- Token fields cross-checked character-by-character against the zai-org GLM-5.3 `chat_template.jinja` (generation prompt, user/assistant/system rendering, turn separator).
- `python3 -m unittest tests.test_data.test_glm_53_template` passes.

This change was developed with AI assistance and verified against the upstream template as described above; happy to adjust per review.


hiteshjain118:
@FrankLeeeee @zyksir @sleepcoo @shuaills , can you please take a look?

PR #895, opened 2026-09-21
fix(eagle3): align loss mask during target preprocessing

### LINKED ISSUE #885: [Bug] Issue with the loss mask of eagle3 in data/parse.py
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/SpecForge/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

In the eagle3 training pipeline, target_head.preprocess shifts BOTH input_ids and target LEFT by one, so at chain step 0 position p carries: 
    input_embed = embed(orig_ids[p + 1])
    aux = aux[p]
    label = argmax(lm_head(orig_hidden[p + 1])) = orig[p + 2]
But loss_mask is NOT shifted — it still marks orig[p] ∈ assistant span. As a result, chain step 0 at position T_start - 1 (which predicts the SECOND assistant token, orig[T_start + 1]) is masked out entirely, even though it's the position that runtime uses in round 1 spec_1. Extending loss_mask backward by 1 recovers loss at that boundary position so the draft is trained to predict "accept vs reject" from the end-of-prompt aux.

Need to change `loss_mask[actual_start:actual_end] = 1` to `loss_mask[max(0, actual_start - 1):actual_end] = 1` in the `parse` function of `GeneralParser`.

### Reproduction

Train command is: specforge train --config xxx
The model is minicpm5-1B

### Environment

h20

navaneethkrishnansuresh:
## Motivation

EAGLE3 shifts `input_ids` and teacher hidden states left during target preprocessing, but previously left `loss_mask` unshifted. This omitted supervision at `T_start - 1` and retained a stale supervised position at the right boundary.

## Modifications

- Shift `loss_mask` together with `input_ids` and `target` in `TargetHead.preprocess`.
- Add regression coverage for boundary, batched, disjoint, boolean, and single-token masks.
- Update compact-teacher and P-EAGLE test doubles to match the production preprocessing contract.

## Related Issues

Fixes #885

## Accuracy Test

An identical clean-`HEAD` versus patched numerical reproduction showed:

- `T_start - 1` gradient: `0.000000` -> `0.113980`
- stale right-boundary gradient: `0.097585` -> `0.000000`
- supervised row count remained unchanged at 3

Local verification:

- 68 relevant EAGLE3, P-EAGLE, feature-loader, and server-capture tests passed
- 3 GPU-only tests skipped locally
- configured Autoflake, isort, Black, and Ruff checks passed for all changed files
- `git diff --check` passed

## Benchmark & Profiling

Not applicable. The change adds one sequence-dimension tensor shift during training preprocessing and does not affect inference.

## Checklist

- [x] Format your code according to the Code Formatting with Pre-Commit guidelines.
- [x] Add unit tests as outlined in the Running Unit Tests guidelines.
- [x] Update documentation / docstrings / example tutorials as needed. No documentation changes are required because this fixes an internal alignment bug without changing public APIs or configuration.
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. Numerical accuracy evidence is included above; performance benchmarking is not applicable.
- [x] For reviewers: no merge-only co-authors were added.


specforge/modeling/target/target_head.py
```diff
@@ -105,8 +105,8 @@ def forward(self, hidden_states):
         return self.fc(hidden_states)
 
     def preprocess(self, input_ids, target, loss_mask):
-        # apply pading
+        # Align each EAGLE row as (x[p + 1], g[p], y[p + 1], mask[p + 1]).
```
curnane-lab: g[p] refers to the teacher/aux hidden state ？

navaneethkrishnansuresh: g[p] refers to the auxiliary hidden state consumed by the draft model while y[p + 1] refers to the final target hidden state projected by TargetHead and for clarity i’ve updated the comment to use the project term

